Issue #19, opened 2018-11-04
Design and prototype monster AI


NathanLovato:
The monsters should be able to target their opponents based on flexible criterias:

- They could try to target the character that's weakest to them
- They could heal once when their life goes below a certain threshold

Etc. I am not sure how to best do this so every option available on the monsters not hardcoded in the base class, without creating a complex gambit system or something like that. Feedback and ideas welcome!

salvob41:
I would gladly take this take of prototyping monster AI. But maybe it would be better, having a getter for the stats' element? 
What do you think about adding chance in taking one or another action (as it is rawly impleemented in PR #24 ) ?

salvob41:
for the healing action, we need to decide how to implement those. See #25

NathanLovato:
v0.2 is already getting a lot of new features, moving this to a future milestone

GresilleSiffle:
It would be useful to have a little visual document like a _decision tree_, or a _behavior tree_ that acts as a reference for the current AI implementation.

Something that even a newcomer can read and figure out how the AI algorithm work.

We can find [examples of them here](https://www.gamedev.net/articles/programming/artificial-intelligence/the-total-beginners-guide-to-game-ai-r4942/).

What do you think about that?

NathanLovato:
And with a flexible system like this, we could build AI for allies as well, have an AI-controlled party member

nhydock:
If you want a nice visual way to do AI composition, I recommend looking at something like Final Fantasy XII's Gambit system.  Not saying we need to have an in-game AI customization feature, but the general structure and templating that gambits provides is a great starting point and pretty flexible.  It's easy to represent enemy AI just as easily with it as character AI.

NathanLovato:
I was thinking about FF XII's gambits with the example above 😄
But I'd still go with a tree structure so we can have multiple conditions, and the benefit of a flexible system for people who'd like to code an RPG with more complex gameplay: e.g. having APs and the ability to use multiple actions per turn.

Dharengo:
I like the idea of a tree as well, but it _may_ lead to a bit of a noob trap where people use complex tree structures to solve a simple problem. What I'm thinking of right now is a gambit-style queue. At the root would be the AI node which doesn't require any input from the developer. The AI node would iterate down its list of children. Lets call these children "Behaviors" for now.

[...]

PR #24, opened 2018-11-06
Prototype monster AI, choose target based on health

salvob41:
As stated in #19 , this is a first implementation of monster AI in which it will choose with a probability `DEFAULT_CHANCE` to attack the player with min health. 

we could add here the possibility to heal itself.

It would be better, maybe, having a getter for the stats?


NathanLovato:
Thanks! You could have clearer variable names and code flow, but it looks like it'll work well. We'll likely need to refactor monster AI from the ground up if the project were to scale a little bit so let's merge for now 😄

Issue #21, opened 2018-11-05
Add monster groups/a way to compose encounters


NathanLovato:
We need some way to create groups of monsters to balance the foes the player will face. It's just lists of monsters to start with (e.g. two slimes and a porcupine), with some code to pick a random group among groups that should spawn in a given area on the map.

Currently there's no notion of formations in the game, front or back rows in combat. I think a basic system will be enough for this project.

nhydock:
I'll take on this task

NathanLovato:
@nhydock is working on this in #28

PR #28, opened 2018-11-18
Add Battle Formations

nhydock:
Addresses #21 by adding a new Node type called Formation.  Extracts out a lot of the spawning logic of battlers into a scene into a factory method.  Formations are used as a form of combat template, and thus also contain meta information such as the kind of environment graphics used when encountering the formation that will also be used when preparing the CombatArena.

Due to how the factory works, also looking into pulling player battlers out of the CombatArena in some way so their stats and presence are persistent across scenes.


godot/combat/CombatArena.tscn
```diff
@@ -129,114 +96,6 @@ mouse_default_cursor_shape = 0
 size_flags_horizontal = 1
 size_flags_vertical = 1
 alignment = 0
-_sections_unfolded = [ "Grow Direction", "Margin" ]
```
NathanLovato: Can't wait for this option to remove unfolded section from the scene files to be available 😄

NathanLovato:
Looking good so far, thanks a lot 🙂

NathanLovato:
Looking great to me, thanks a lot for your work! Thanks to you, Guilherme, and saalazar, I could spend some time working on art and pimping up the project :)

nhydock:
two Party.gd files got committed accidentally.  The one that's in the root godot folder shouldn't exist

NathanLovato:
Fixed, thanks for the heads-up :)

Issue #15, opened 2018-11-04
Add a base class to represent Magic and/or Skills


NathanLovato:
Create a class that we can use to store all the data related to a special attack a battler can use.

Should this class extend Resource or Reference? We'd use it to store only the data related to a skill (then another Object or Node could use the resource and provide functionality on the Battler, or render it in the UI).
See `StartingStats` https://github.com/GDquest/godot-turn-based-rpg/blob/master/godot/combat/battlers/jobs/StartingStats.gd

guilhermehto:
This was closed through #32 I believe.

PR #32, opened 2018-11-19
Skills system added

### LINKED ISSUE #16: Add a Magic or Skill CombatAction
Currently the characters can only attack, because they only have this action available. This task is about adding support for using Magic in Combat, or from a menu. See #15  for the data part of this task

guilhermehto:
It's now possible to add skills as resources and use them in combat.
Some changes:

- Initial stats now have mana
- A combat action for skills has been added
- A really simple menu to select which action to perform in the given
turn
- Resource for skills
- Skills can be executed on a given chance

What can be improved:
- As of now, the skills have a `base_damage` which is added on top of the
already calculated damage in the `Hit` script. We might want to use the
strength as a multiplier to increase the overall damage of the skill,
for instance, instead of just adding them together.

- The menu for selecting the action is really simple and the UX is bad
to say the least. I'll create a separate issue for this.

- Skills are only available as a form of attack, i.e. you can't create
healing skill with the current system. But it can be extended to do so.

- They are not specific to jobs. For instance, both Dan and Blue are
fighters, but you'll have to add skills to them separately. Though, this
is most due to how they were created. We should probably create a
separate FighterJob and add the skills to them, this would fix the
problem imo.

Closes #16 and #15

### DIFF
--- a/godot/combat/battlers/skills/CharacterSkill.gd
+++ b/godot/combat/battlers/skills/CharacterSkill.gd
@@ -1,0 +1,15 @@
+extends Node
+
+class_name CharacterSkill
+
+var skill_description : String
+var mana_cost : int
+var base_damage : int
+var success_chance : float
+
+func initialize(skill : Skill):
+	name = skill.skill_name
+	skill_description = skill.skill_description
+	mana_cost = skill.mana_cost
+	base_damage = skill.base_damage
+	success_chance = skill.success_chance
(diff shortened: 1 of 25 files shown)


godot/combat/battlers/skills/CharacterSkill.gd
```diff
@@ -0,0 +1,15 @@
...
+
+class_name CharacterSkill
+
+var skill_description : String
+var mana_cost : int
+var base_damage : int
+var success_chance : float
+
+func initialize(skill : Skill):
+	name = skill.skill_name
```
NathanLovato: When you use a resource, it's already an object with member variables - no need to duplicate them

PR #34, opened 2018-11-20
Add Character Growth

nhydock:
Characters gain experience points at the end of battle, which contribute to their level.  Levels determine stat growth along defined curves for each character.

Stats now have an experience value required.  In the case of characters, this determines their level.  In the case of enemies, it determines how much they reward.

Post battle a rewards popup will display what each surviving character got in earned experience, as well as any level up messages.  The popup is also prepped for adding item drop messages when we get around to adding it.


NathanLovato:
Thank you so much for your work :)

PR #35, opened 2018-11-20
Mana bars added

### LINKED ISSUE #31: Add mana bar
With the implementation of skills - see #16 - battlers will now have mana to spend on skills. Thus, the mana should also be shown in the UI, below the life bar.

guilhermehto:
The bars system has been refactored. It's now possible to easily extend
it to create different bars i.e. stamina, rage, energy, etc..
The interface anchor has been removed and replaced with a proper control
node to hold the bars. Folder structure has also been cleaned.

Note that the newly added manabar will not show as of now since the
characters do not have mana at level 1 (not sure why). They were
configured to only appear if the battler has more than 0 of max mana.

Closes #31


godot/combat/interface/bars/lifebar/LifeBar.gd
```diff
@@ -0,0 +1,12 @@
+""" Can't extend from StatBar (the class_name) because this won't let us export
+the inherited variables from the base class """
+extends "res://combat/interface/bars/StatBar.gd"
+class_name LifeBar
+
+func _connect_value_signals(battler : Battler) -> void:
+	var battler_stats = battler.stats
```
NathanLovato: A detail, but it's okay to access battler.stats both times

godot/combat/interface/bars/manabar/ManaBar.gd
```diff
@@ -0,0 +1,12 @@
+""" Can't extend from StatBar (the class_name) because this won't let us export
+the inherited variables from the base class """
+extends "res://combat/interface/bars/StatBar.gd"
+class_name ManaBar
+
+func _connect_value_signals(battler : Battler) -> void:
```
NathanLovato: Fine for now, but maybe there's a better way to handle life and mana than having duplicated code like that - e.g. tracking the property automatically in some way by identifier (as they're all named max_*, *, *_changed, *_depleted)

NathanLovato:
Looking good 😄

NathanLovato:
You can add a screenshot when you make visual changes like these, just to show what your PR achieves

PR #36, opened 2018-11-21
Add dialogue system to LocalMap

### LINKED ISSUE #22: Map: integrate the conversation system
Also create an NPC that takes one cell and that the player can interact with. The job is mostly to port your work on the grid-based & conversation demo to Godot 3.1, integrating it in here.

henriiquecampos:
With this I think we close #22 I didn't create a NPC, need feedback on the implementation first, so what I did was just create a basic dialogue that appears when you pick an item.

I made some changes to what we have in the RPG demo in Godot, there we have a DialoguePlayer for each think that could potentially start a Dialogue, now we have Dialogues, which parse JSON files into Dictionaries when Dialogue.load_dialogue() and they have the path to the JSON file.

Now the DialoguePlayer is a node responsible only for keep and communicate the flow of the Dialogue, if it started, ended, what is the current data (name, text) etc...

With this we instead of having to grab DialoguePlayer nodes around the memory, we just keep track of the Dictionary.

I did this because the Object was getting free and we lost reference to its DialoguePlayer, with this current approach it just passes the data we need and can be free, no worries.

### DIFF
--- a/godot/dialogue/Dialogue.gd
+++ b/godot/dialogue/Dialogue.gd
@@ -1,0 +1,19 @@
+extends Node
+
+export (String, FILE, "*.json") var dialogue_file : String
+
+func load_dialogue() -> Dictionary:
+	"""
+	Parse a JSON file and returns it or an empty dictionary if the
+	file doesn't exist.
+	"""
+	
+	var file = File.new()
+	
+	if file.file_exists(dialogue_file):
+		file.open(dialogue_file, file.READ)
+		
+		var dialogue = parse_json(file.get_as_text())
+		
+		return dialogue
+	return {}
--- a/godot/dialogue/Dialogue.gd
+++ b/godot/dialogue/Dialogue.gd
@@ -1,8 +1,8 @@
 extends Node
 
-export (String, FILE, "*.json") var dialogue_file : String
+export (String, FILE, "*.json") var file_path : String
 
-func load_dialogue() -> Dictionary:
+func load() -> Dictionary:
 	"""
 	Parse a JSON file and returns it or an empty dictionary if the
 	file doesn't exist.
@@ -10,10 +10,9 @@ func load_dialogue() -> Dictionary:
 	
 	var file = File.new()
 	
-	if file.file_exists(dialogue_file):
-		file.open(dialogue_file, file.READ)
-		
+	if file.file_exists(file_path):
+		file.open(file_path, file.READ)
 		var dialogue = parse_json(file.get_as_text())
-		
 		return dialogue
+	
 	return {}
--- a/godot/local_map/DialogueBox.gd
+++ b/godot/local_map/DialogueBox.gd
@@ -13,7 +13,7 @@ func initialize(dialogue):
 	next_button.grab_focus()
 	next_button.text = "Next"
 	
-	dialogue_player.start_dialogue(dialogue)
+	dialogue_player.start(dialogue)
 	
 	update_content()
 	show()
@@ -26,7 +26,7 @@ func _on_dialogue_player_finished():
 	hide()
 
 func _on_next_button_down():
-	dialogue_player.next_dialogue()
+	dialogue_player.next()
 	update_content()
 	
 func update_content() -> void:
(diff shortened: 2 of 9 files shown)


henriiquecampos:
Ohh btw, this also allow to pass a dictionary to a DialoguePlayer and use in the CombatSystem, so we can communicate damage, skill usage, etc...

godot/local_map/grid/grid.gd
```diff
@@ -24,6 +24,11 @@ func request_move(pawn, direction):
 			return update_pawn_position(pawn, cell_start, cell_target)
 		CELL_TYPES.OBJECT:
 			var object_pawn = get_cell_pawn(cell_target)
+			
+			if object_pawn.has_node("Dialogue"):
```
NathanLovato: That's not a good idea. The grid should let the pawn itself decide on the interaction. It's fine to have it like this in this PR, for testing, but as the game's complexity increases I don't want us to handle all possible interactions between characters on the grid node.

godot/local_map/DialogueBox.gd
```diff
@@ -0,0 +1,34 @@
...
+	
+	yield(next_button, "button_down")
+	
+	hide()
+
+func _on_next_button_down():
+	dialogue_player.next_dialogue()
+	set_dialogue()
+	
+func set_dialogue() -> void:
```
NathanLovato: This function isn't setting the dialogue. It updates the dialogue box's labels, doesn't it?

godot/local_map/DialogueBox.gd
```diff
@@ -0,0 +1,34 @@
...
+onready var dialogue_player : DialoguePlayer = get_node("DialoguePlayer")
+onready var name_label : Label = get_node("Panel/Colums/Name")
+onready var text_label : Label = get_node("Panel/Colums/Text")
+onready var next_button : Button = get_node("Panel/Colums/Next")
+
+func _ready():
+	next_button.connect("button_down", self, "_on_next_button_down")
+	dialogue_player.connect("dialogue_finished", self, "_on_dialogue_player_finished")
+	
+func _on_LocalMap_dialogue_started(dialogue):
```
NathanLovato: This shouldn't be a specific callback. This function kicks off the conversation. How about `start`?

godot/dialogue/Dialogue.gd
```diff
@@ -0,0 +1,19 @@
+extends Node
+
+export (String, FILE, "*.json") var dialogue_file : String
+
+func load_dialogue() -> Dictionary:
```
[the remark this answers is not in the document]
henriiquecampos: I implemented this on ffab5c2054b0dfdad9246915c039a20114aac39a, it works just OK, but I'm afraid it can be confusing later on, as it is highlighted since `load()` is also a GDScript built-in function.

Can we keep it that way or maybe find another name?

godot/dialogue/Dialogue.gd
```diff
@@ -0,0 +1,19 @@
+extends Node
+
+export (String, FILE, "*.json") var dialogue_file : String
```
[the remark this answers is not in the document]
henriiquecampos: I used `file_path` instead, since it doesn't turn out to be an actual file, so `file` could lead to think that this is the actual `File`, as if a `File` class instance.

godot/local_map/grid/grid.gd
henriiquecampos: Shall we move these, together with [these ](https://github.com/GDquest/godot-turn-based-rpg/blob/349adf09fd7601ae31f1bf81cd95a80dd0927632/godot/local_map/grid/grid.gd#L34-L36) to a function on `pawn.gd` or something?

NathanLovato: Let's do that later, once we add actual content

godot/Game.gd
```diff
@@ -18,7 +18,7 @@ func enter_battle(formation: Formation):
 	"""
 	if transitioning:
 		return
-
```
NathanLovato: Can you remove changes to this file?

godot/Party.gd
```diff
@@ -0,0 +1,5 @@
+extends Node
```
NathanLovato: Also this one shouldn't be here. I just removed it from master. It seems like you rebased from your fork and not this repository

Issue #37, opened 2018-11-21
Map: Prevent Player from moving on the map during battle


godofgrunts:
Currently, during battle, the actor can freely move around the map. This allows the player to initiate multiple battles at the same time and causes a disconnect since the player appears to have randomly teleported while in battle.

henriiquecampos:
I think we can make something like in the Godot rpg-demo, where we remove the map from the process tree, than we insert it again when the combat is over

https://github.com/godotengine/godot-demo-projects/blob/master/2d/role_playing_game/Game.gd#L10-L20

https://github.com/godotengine/godot-demo-projects/blob/master/2d/role_playing_game/Game.gd#L39-L41

Any suggestions on how to handle that @NathanLovato ?

NathanLovato:
closed with #38

PR #38, opened 2018-11-21
Changed local_map.visible to add/remove_child to prevent player movem…

godofgrunts:
With Pigdev's help, I edited game.gd to change the local_map.visible to an add_child and remove_child.

This prevents the players from moving during battle. #37


NathanLovato:
Great, thanks to the both of you 😄

PR #39, opened 2018-11-22
Implement minimal Attack and UseSkill CombatAction

### LINKED ISSUE #25: Implementing Actions
This is an issue to design and implement different actions, such as Healing, Magic, Item, Special (?). 

We could do as it is now for the Attack. Nodes that implements actions.

### LINKED ISSUE #33: Improve actions menu
This is the current look of the actions "menu":

image

The new menu shall:
- [ ] Show the mana cost for actions
- [ ] Action's description
- [ ] Action's base damage
- [x] Prevent actions from being selected if not enough mana available
- [x] Better UX overall (bigger text, only visible when necessary, etc.)

MarianoGnu:
Given the feedback from everyone i changed the PR to make it much more easy to read and write than before. All routines are handled by GdScriptFunctionState now instead of custom signals

- To create a new subroutine just add a function to the CombatAction.gd script
- To create a new CombatAction:
  1) Extend CombatAction
  2) implement the _internal_execute function making sure you yield on subroutines when you want to wait them to finish before continue with the next step.

*Original Post*
> This is a very minimal application of the Command Pattern, wich allows setup complex commands like Attack, Defend, Skill, Magic, Use Item, Flee and other commands being implemented using small and easy to create commands. Closes #25 
> 
> To create a new Command:
> 1) Extend BattleActionCommand.gd
> 2) Override the execute command
> 3) when you finish, emit the "execute_finished" signal, making sure at least one frame has passed since the function was called (otherwise yield function wont be able to catch the signal)
> 
> To create a new CombatAction:
> 1) Extend CombatAction.gd
> 2) override the setup(targets) function.
> 2.1) in this function use the push_command(command) function to add a new function this makes sure the command is properly initialized.
> 
> Also i have improved the ActionList User Interface. Closes #33 
> 
> I could have fixed other issues without realizing, found some problems while implementing the Commands.


godot/combat/battle_action_command/AttackTargetCommand.gd
```diff
@@ -0,0 +1,12 @@
+extends BattleActionCommand
+
+class_name AttackTargetCommand
+
+func execute():
+	#Always yield for a frame, that way signal can be catched using yield
+	yield(action.actor.get_tree(), "idle_frame")
+	
+	action.actor.attack(action.target)
+	yield(action.actor.get_tree().create_timer(1.0), "timeout")
```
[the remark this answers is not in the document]
MarianoGnu: This is because the attack function is not a routine, it doesnt play animations(yet) or waits for signal or time (also i copypasted this timeout from somewhere else)

godot/combat/battle_action_command/BattleActionCommand.gd
```diff
@@ -0,0 +1,15 @@
...
+
+class_name BattleActionCommand
+
+signal execute_finished;
+
+var action = null
+
+func execute():
+	#Always yield for a frame, that way signal can be catched using yield
+	yield(action.actor.get_tree(), "idle_frame")
```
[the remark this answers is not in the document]
MarianoGnu: Example:
```
func routine():
    subroutine()
    yield(self, "custom_signal")

func subroutine():
    emit_signal("custom_signal")
```

By the time the signal is emmited, the routine function didn't reach the yield statement. This happened to me in the past with MoveCommands when the actor was already on the desired position

MarianoGnu:
Thx for feedback, sadly i'm already on bed and changes will have to wait till tomorrow, unless you decide to merge it and fix it yourself.

godot/combat/battle_action_command/BattleActionCommand.gd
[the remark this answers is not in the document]
NathanLovato: Sure, but can't you do `yield(self.subroutine, "completed")` ? Then you let the subroutine yield on a timer, an AnimationPlayer, etc.

NathanLovato:
It can wait, thanks :)

godot/combat/battle_action_command/BattleActionCommand.gd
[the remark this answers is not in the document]
MarianoGnu: You could change CombatAntion.execute to work that way, dont remember if i had a reason to use a signal, but in the future we will want to allow some kind of cancel, to allow the player use an skill instead of attack

godot/project.godot
```diff
@@ -9,6 +9,16 @@
 config_version=4
 
 _global_script_classes=[ {
+"base": "Reference",
+"class": "AttackTargetCommand",
```
nhydock: Considering specific command types most likely won't ever be extended or their type used globally, I don't feel like it's good to have them be registered classes.

godot/combat/battlers/Battler.gd
```diff
@@ -50,6 +48,11 @@ func attack(target : Battler):
 	var hit = Hit.new(stats.strength)
 	target.take_damage(hit)
 
+func can_use_skill(skill : Skill) -> bool:
+	if stats.mana < skill.mana_cost:
```
nhydock: return this conditional in a single line instead of breaking into return false and true statements

godot/combat/battlers/actions/Attack.gd
```diff
@@ -1,5 +1,19 @@
...
+	targets = p_targets
+	actor = get_node("../..")
+	command_queue.clear()
+	print("TODO: ATTACK -> at some point we should define the attack commands based on the equiped items\n" +
+		  "ie: if you have a bow you should not move to the target!")
+	if actor.party_member:
+		push_command(SelectTargetCommand.new())
+	else:
+		pass
+	push_command(MoveToTargetCommand.new())
```
nhydock: I'm personally not a fan of the sequential list of commands vs just doing them explicitly in the execute method.  This prevents you from having certain commands be okay to execute in parallel while others need to be yielded in sequence.  Also it doesn't really save anything in terms of memory since the commands are RefCounted and will be cleared after use if just declared and used locally within execute.

godot/combat/interface/SelectArrow.gd
```diff
@@ -35,10 +37,10 @@ func move_to(battler : Battler):
 		Tween.EASE_OUT)
 	tween.start()
 
-func _unhandled_input(event):
-	if not visible:
+func _input(event):
```
nhydock: typically for gameplay input you want to use _unhandled_input (as detailed in the [Godot docs tutorials](http://docs.godotengine.org/en/3.0/tutorials/inputs/inputevent.html)).  I'd probably just make another "enabled" flag instead of toggling the actual input processor for the node.

godot/project.godot
MarianoGnu: They are used for readabylity when they are created on the setup function

godot/combat/interface/SelectArrow.gd
MarianoGnu: Using unhandled input makes they ui capture the input and this code never runs

godot/combat/battlers/actions/Attack.gd
MarianoGnu: This is done this way for readability and reusage, for example this code implements a Monster AI and is very easy to read and edit 
```
if actor.status.health < actor.stats.max_health / 10:
    if actor.ai_can_heal() and randi()%100 < 50:
        push_command(AiHealSelfCommand.new())
    elif randi()%100 < 10:
        push_command(AiEscapeCommand.new())
    else:
        push_attack
else:
    push_attack()

func push_attack():
    if randi()%100 < 50:
        push_command (AiSelectWeakestTarget.new())
    else:
        push_command (AiSelectRandomTarget.new())
    push_command (MotoToTargetCommand.new())
    push_command (AttackCommand.new())
    push_command (ReturnToStartPositionCommand.new())
```

godot/combat/battle_action_command/BattleActionCommand.gd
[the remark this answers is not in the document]
NathanLovato: Yup but I don't think the ability to cancel requires waiting a frame. It'll be handled on the UI's side - the user can go up to the target selection and go back, at this point the battle system hasn't executed the action/commands yet

godot/combat/battlers/actions/Attack.gd
NathanLovato: I couldn't tell which approach would work best until we have actual gameplay at this point. Like @nhydock I like the idea of keeping the code as simple as possible, calling methods directly on the battler. But I don't know how either approach scales and a command queue also makes sense.

godot/combat/interface/SelectArrow.gd
NathanLovato: Which UI captures it? The command select menu maybe? If so it means it doesn't properly close. You shouldn't have to change the method to _input, true

guilhermehto: Probably the actions menu, it's just an `ItemList` atm. When you are selecting targets, the item list also changes the selected action. This will get fixed in #33

guilhermehto:
Hey, thanks for your work! 

There are 3 missing functionalities needed for the actions menu, this PR doesn't close #33 
You can check which ones I'm talking about in the issue itself.

godot/combat/interface/SelectArrow.gd
MarianoGnu: Probles is that i want the ActionList visible,but it was not way to disable input, what i did was capturong the focus with a button out of screen, but button try to navigate using ui_dir buttons.
Other way could be replace ui_direction actions for custom actions, that way gui wont capture them

godot/combat/battlers/actions/Attack.gd
nhydock: The more I fiddled with this code the less I like it.  Command pattern is overkill for breaking up reusable acts within an Action.  Typically the code that's reusable is only a few lines long and will only be reused in 2-3 places, so it's fine to have it duplicated.  The amount of work to wrap that in a command and the overhead of how actions are now prepared with commands in mind actually makes it counterproductive.

MarianoGnu: actually, i'd usually write the commands as single functions on the CombatActionand then extend it to overwrite the execute method, but i thought this looks more the way Nathan usually programs stuff, since he preffers small objects with very specific functionality, and if posible visible on the scene tree view, so you can have an easy look to what it does just looking at the hierarchy. Here we have not hierarchy, but the setup method speaks by itself (more or less)

MarianoGnu:
@NathanLovato @nhydock @guilhermehto
I Rewrote how the subroutines are handled to simplify how creation of new Commands, and you where right, it looks much more clean now and we dont have a lot of 2 or 5 lines scripts

godot/combat/interface/SelectArrow.gd
NathanLovato: Two other options:
- Make the arrow a Control node that uses _gui_input, and make it grab focus when it becomes active
- Toggle process_gui_input off on the itemlist if you want to keep it on screen

It may be a job for another PR though.

godot/combat/CombatAction.gd
```diff
@@ -2,11 +2,49 @@ extends Node
...
+func execute(p_targets : Array, p_allys : Array) -> void:
+	targets = p_targets
+	allys = p_allys
+	yield(_internal_execute(), "completed")
+
+func _internal_execute() -> void:
 	print("%s missing overwrite of the execute method" % name)
-	emit_signal("execute_finished")
+
+func attack_routine():
```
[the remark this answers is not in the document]
MarianoGnu: Sorry but i wont change this again. take it or leave it.
Supose you implement use_skill_routine on SkillAction and use_item_routine on ItemAction, then if you need to create something like a ItemSkillAction (a consumible item that triggers an skill, like throwing a molotove bomb) then you wont be able to use both.
thaving these routines on the base class makes sure it's available anywhere, and if oyu need veri specific variants on an action you can always override them.

MarianoGnu:
@NathanLovato updated with the requested changes.
- removed _internal_execute call
- added an initialize call
- actor initialization now happens on the initialize call
- SelectArrow now is a control, wich grabs the focus
- replaced _input for _gui_input

NathanLovato:
Thanks! Let's merge and keep the project going now :)

Issue #20, opened 2018-11-04
Write a changelog for v0.1 and v0.2


NathanLovato:
Like https://github.com/GDquest/Blender-power-sequencer/blob/master/changelog.md, to keep track of the project's growth

godofgrunts:
You probably should tag releases if you want to do this. I have no idea where 0.1 ended and 0.2 began.

PR #40, opened 2018-11-22
Added CHANGELOG.md by following git commits.

godofgrunts:
I found a site that used what I consider sane standards for a [change log](https://keepachangelog.com/en/0.3.0/) and decided to go with that for now since @NathanLovato was not on Discord at the time.

I did my best to compress multiple changes into just one line if possible.

If I may be so bold, I would recommend changing the contributor guidelines to say that your PRs must include and entry in the CHANGLELOG.md under the [Unreleased] section. This way the change log is always up to date.

This is in reference to #20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,43 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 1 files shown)


NathanLovato:
Thank you very much 🙂

I'm going to fill 0.1 right away

PR #48, opened 2018-11-22
clean up .import folder

salvob41:
Welll... This PR has a low priority and it is a bit OCD :D But.. here it is :D


.gitignore
```diff
@@ -1 +1 @@
-.import
+godot/.import
```
NathanLovato: No need for that, .import will exclude any folder named `.import`

NathanLovato:
Done manually in  97c8412

PR #49, opened 2018-11-23
[WIP] Add Leader-Follower behavior

### LINKED ISSUE #23: Map: Following the player
Code a pawn/character that can follow the player. As it's grid-based motion, it just needs to reproduce the player's movement sequence with a one, two, three steps delay (e.g. if there's 4 party members - it's like the game snake). 

This pawn shouldn't collide with the player. Also look at the Battler and BattlerAnim/BattlerSkin(?) scenes: I want us to always have the animated character skins separate from the controllers themselves so that we can swap them anytime, and later work in parallel with artists.

I'm not sure how to best track the player's motion. I would use a signal, e.g. `moved(direction : Vector2)`, and store the last X moves in an array/stack/heap, erasing them as you go. Feel free to suggest something better!

You can use a YSort node to sort all the characters on the map. Don't worry about conflicts if both pawns end up on the same cell for now.

henriiquecampos:
To achieve this in a more scalable way I had to refactor the current implementation of the `Actor.gd`, currently it was also responsible to handle inputs, since both a Leader and a Follower are actors in the LocalMap context, they shouldn't both handle inputs, so instead now only the `Leader.gd` handles them, leading the others based on the direction the Leader handled.

I also renamed the files to fit our current convention where classes and scenes have **PascalCase**.

With this we just need to define the YSort architecture to close #23


NathanLovato:
In the latest master you can nest YSort nodes now, but for now you can just have the characters in a YSort along with the monsters. We have yet to define art on the map.

henriiquecampos:
Well fug, I think my internet trolled me, I was sure that I mentioned the trouble with the YSort insertion in this system in #23 but it doesn't seem to be there.

The problem is that the YSort needs to be the parent of the `Pawns` so it kinda change the structure of the system by a bit, making the Grid reference to the `YSort` so it can properly get the `Pawns` and also make the `YSort` become sort of a `PawnContainer` that delegates the `Pawn` requests to move.

This is the simplest architecture I could find, I'd much appreciate a better way to approach that, especially by not coupling the `Grid` to a `PawnContainer`

But if this current one is OK, I thing we are good to merge

Issue #47, opened 2018-11-22
Player inventory


guilhermehto:
A menu to display the items that the player currently has with him.

I believe that it shouldn't be complex, just a list should do the trick.

Issue #46, opened 2018-11-22
Equipable items


guilhermehto:
Pieces of equipment for players.

Should they also restrict which jobs can equip them?

Issue #45, opened 2018-11-22
Base class for items


guilhermehto:
A base class for any type of item (usable, equipable, etc..)

Following what we've been doing so far, I believe the data should be separate from functionality aka use resources for them.

Issue #44, opened 2018-11-22
Usable items


guilhermehto:
This issue is about creating a base system for usable items, i.e. healing potions, attack potions, food, etc..

nhydock:
I haven't looked at the ARPG demo yet, but I'm still interested in working on the inventory management system and tying it to commands.  With items I also feel that it'll be important to further improve actions to allow for different in-menu and in-combat behavior.

PR #52, opened 2018-11-26
Add basic item and inventory handling

nhydock:
This adds a managed Inventory that sits attached to the party node.  Inventory manages both the possession of items as well as currency which may be used in shops to purchase other items that may not be found as drops on enemies.

Items are defined as resources, with basic items acting as purely loot that may be sold for currency, or as keys.  Extended off of items are both Equipment and Consumables.  Equipment items may have stats which stack on top of a player battler's normal growth.  Consumables may have commands defined on them so they may be used in combat.

This tackles the base work for #44 #45 #46 and #47


NathanLovato:
Looking good, thanks once again 🙂

PR #58, opened 2018-11-26
Added the ability to cancel the selection arrow

### LINKED ISSUE #54: Cannot go back from the select arrow to the CombatAction menu
Add ability to cancel enemy selection and to go back to the command select menu

guilhermehto:
Had to remove the typed part of the `execute()` function from CombatAction.gd. It wasn't a `void` anymore: it returns if it actually managed to hit a target. When changing the function type to bool, yields would throw an error. I'm not sure if this is a bug with typed GDScript or what.

Closes #54


NathanLovato:
Crashing when you yield to a bool sounds like a bug yes

PR #60, opened 2018-11-26
Added ability to press ui_cancel during attack to cancel

godofgrunts:
This is a fix for #54 

Because there are so many yields and, as far as I can tell, you can't break out of yields cleanly I ended up returning `null` to to `target_selected` in `SelectArrow` and followed it all the way back to the CombatAction.

I'm fairly new to git and I don't know how to prevent the .tscn files from being commited since I didn't actually make any changes to them.


NathanLovato:
Thanks a lot! As Gui also tackled the issue and his PR was (you pushed default_env.tres) cleaner I merged his.
I'm glad to have you around to give a hand 🙂
Be sure to comment on issues you're working on, and you can get in touch if you want to work on issues that aren't open yet!

PR #61, opened 2018-11-26
Added monster drops

### LINKED ISSUE #59: Drops from monsters
Allow the user to configure the possible dropped items from monsters and their rates.

guilhermehto:
Monsters can now be configured to drop items based on a resource (MonsterDrops.gd).
The usability is not the greatest, but that's due to not being able to export custom types. 
There were two possible approaches that I saw:
1. An `Array` holding dictionaries for each item that can be dropped along with the data needed for the drop behavior.
2.  Create a custom resource that would encapsulate all the data for the items that can be dropped from a monster: The item itself, min and max quantities, and the chance. 

I decided to go with the first one.

ps: The items are _NOT_ being added to the player's inventory . This is just the functionality needed for the monsters to start dropping items.

Closes #59


godot/combat/battlers/enemies/porcupine/PorcupineDrops.tres
```diff
@@ -0,0 +1,15 @@
+[gd_resource type="Resource" load_steps=3 format=2]
```
nhydock: instead of making this another file, might we consider having an embedded resource?  Something like drop rates typically wouldn't be shared and require saving as another file.

godot/combat/battlers/enemies/MonsterDrops.gd
```diff
@@ -0,0 +1,20 @@
...
+		'chance': 0.5
+	}, {
+		'item': Potion.tres,
+		'min_quantity': 3,
+		'max_quantity': 20,
+		'chance': 0.25
+	}, 
+"""
+
+export var items : Array
```
NathanLovato: @nhydock Any suggestion to improve the editing of the drops list through the inspector? How do you handle this in your project? Having to type dicts manually isn't ideal. Right now all I can think of is coding an inspector plugin that's specific to our items.

godot/combat/Rewards.gd
```diff
@@ -21,7 +22,15 @@ func initialize(battlers : Array):
 	
 func _add_reward(battler : Battler):
 	experience_earned += battler.stats.experience
-	# TODO add item drops
+	if battler.drops != null:
+		for drop in battler.drops:
+			if drop.chance < 1.0 and rand_range(0.0, 1.0) > drop.chance:
```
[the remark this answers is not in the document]
nhydock: instead of rand_range use randf

godot/combat/Rewards.gd
```diff
@@ -56,7 +65,7 @@ func on_battle_completed():
 		$Panel/Label.text = "%s Leveled Up to %d" % [battler.name, battler.stats.level + 1]
 		yield(get_tree().create_timer(2.0), "timeout")
 	for drop in drops:
-		$Panel/Label.text = "Found %s" % drop.name
+		$Panel/Label.text = "Found %s %s(s)" % [drop.quantity, drop.item.name]
```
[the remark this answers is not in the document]
nhydock: this follows convention of final fantasy for the SNES, where it will show each different kind of item one at a time.  It's looping through and yielding.  Follows the convention I set earlier with the reward UI for showing EXP and Level ups

godot/combat/battlers/enemies/MonsterDrops.gd
nhydock: This is pretty similar to my game as well, sadly.  It's definitely not the best of my inspector accessible data types.  The alternative was to make Drop be a resource type instead of Dict and embed them.  Another alternative is make the entire drop logic be embedded resources that extend a Drop class, then override a function.  A generic extension would be what we have here, but we could open the doors for battles where you can get different reward for different accomplishments, such as beating a boss in under 5 minutes. 

Something we might want to consider, if we add in the RPG traditional steal command to make that function properly, monsters should have their own inventories.  So then if you steal and item from them they won't drop it later.

guilhermehto: I was thinking about a tool as well, but the best solution would probably be letting enemies have their own inventory, as @nhydock suggested.

nhydock: Even if we just give them a simple inventory, there still needs to be a data type that identifies drop rates so we know which item to give.

Issue #70, opened 2018-11-29
Animated number or icon (for damage, healing, status... in combat)


godofgrunts:
In a lot of JRPGs, when a unit is attacked, the damage that was done would appear giving the player an easy to digest number (as opposed to doing mental math on the spot). This is also done for healing, status effects, etc.

I think this would be a good addition.

As a mock-up I made:

![Pop up gif](https://giant.gfycat.com/MinorSereneBeetle.gif)

Speaking with faildozer on Discord, he suggested adding a canvas layer for these popups and emitting a signal to the canvas layer which I think is a great idea.

godofgrunts:
I'm going to try to implement this tomorrow if I have some down time.

PR #71, opened 2018-11-29
[WIP] Added Popup information with font. #70 

godofgrunts:
This PR is in reference to #70 

I've started working adding the popups. In the example below you can see damage occurring. I'm going to also do Mana as well.

I would like some advice on positioning though.

Right now I'm doing  `rect_position = Vector2(battler.position.x, battler.position.y - 150)` but 150 is just an arbitrary number and doesn't look nice on the players. 

popups


godot/combat/interface/PopUp.gd
```diff
@@ -0,0 +1,19 @@
...
+	for battler in battlers:
+		_connect_value_signals(battler)
+	
+func _connect_value_signals(battler : Battler) -> void:
+	var battler_stats = battler.stats
+	battler_stats.connect("health_changed", self, "_on_value_changed", [battler])
+
+func _on_value_changed(new_value, old_value, battler) -> void:
+	var value = new_value - old_value
+	rect_position = Vector2(battler.position.x, battler.position.y - 150)
```
nhydock: you could get the sprite on the battler, then use the sprite's rectangle to offset the label to a position relative to the battler, such as positioning it at their feet.

.gitignore
```diff
@@ -1,3 +1,3 @@
-.import
+*.import
```
NathanLovato: No, you shouldn't change that, and you shouldn't never ignore the *.import files: they contain import metadata/flags
On the other hand the .import folder contains cached copies of imported images and resources that are local to your computer so you don't want them in the repo.

godot/assets/fonts/munro.ttf.LICENSE
```diff
@@ -0,0 +1,43 @@
+SIL Open Font License
```
NathanLovato: Is there no alternative to having a license file per font? I'd rather not have extra files hanging around the repo.

godot/combat/battlers/characters/Godette.tres
```diff
@@ -11,4 +11,5 @@ script = ExtResource( 2 )
 anim = ExtResource( 1 )
 stats = ExtResource( 4 )
 skills = [ ExtResource( 3 ) ]
+drops = null
```
NathanLovato: This shouldn't be in this PR, and the drops shouldn't initialize to `null`

godot/combat/interface/PopUp.gd
```diff
@@ -0,0 +1,19 @@
+extends Control
+
+func initialize(battlers : Array) -> void:
+	for battler in battlers:
+		_connect_value_signals(battler)
+	
+func _connect_value_signals(battler : Battler) -> void:
+	var battler_stats = battler.stats
+	battler_stats.connect("health_changed", self, "_on_value_changed", [battler])
```
NathanLovato: Favor `_on_NodeName_signal_name` for your callback methods: `_on_Battler_health_changed`

godot/combat/interface/PopUp.gd
```diff
@@ -0,0 +1,19 @@
+extends Control
+
+func initialize(battlers : Array) -> void:
+	for battler in battlers:
+		_connect_value_signals(battler)
```
NathanLovato: No need for a separate method:
```
func initialize(battlers : Array) -> void:
	for battler in battlers:
		battler.stats.connect("health_changed", self, "_on_value_changed", [battler])
```

godot/combat/interface/PopUp.gd
```diff
@@ -0,0 +1,19 @@
...
+	
+func _connect_value_signals(battler : Battler) -> void:
+	var battler_stats = battler.stats
+	battler_stats.connect("health_changed", self, "_on_value_changed", [battler])
+
+func _on_value_changed(new_value, old_value, battler) -> void:
+	var value = new_value - old_value
+	rect_position = Vector2(battler.position.x, battler.position.y - 150)
+	$Label.text = str(value)
```
NathanLovato: I prefer to have `onready var label = $Label` at the top (same for other nodes) so that:

- The reference to the nodes a class gets is always cached. Getting many nodes throughout your codebase can be costly
- There's only one place to change node paths, and it's always at the top of the gdscript files
- at a glance, opening any file, you know which nodes they access

godot/assets/fonts/munro.ttf.LICENSE
godofgrunts: We don't even have to use this font to be honest. I just didn't want to have to mess with scaling and stuff when I was working on the popup and Godot doesn't allow me to change the font size of the default font for whatever reason.

As far as having alternative licenses, we could just pick a license and only use fonts with that license I suppose. I'll leave that up to you though.

godot/combat/battlers/characters/Godette.tres
godofgrunts: Yeah I didn't manually change that, but I wasn't sure if something I did in the code changed Godette and Robi. I'll make sure to not commit those in the future.

godot/combat/interface/PopUp.gd
godofgrunts: Originally I was going to use the one function to do all popups. Do you think, for the sake of clarity, that I should have multiple functions with more specific names? 

EG.

_on_Battler_health_changed
_on_Battler_mana_changed
_on_Battler_status_changed

etc

godot/combat/interface/PopUp.gd
godofgrunts: Makes perfect sense. Thanks.

.gitignore
godofgrunts: Oh I didn't know that that import contained import metadata stuff. Thanks.

godot/combat/interface/PopUp.gd
godofgrunts: Of course now that you point it out it makes perfect sense.

godot/combat/interface/PopUp.gd
godofgrunts: As discussed in discord, I don't think there is a good way to do this right now. Leaving with a number for now.

godofgrunts:
I've pushed some more changes to this branch (I hope I did this right, I had to rebase because of the changes made upstream).

I ended up changing quite a bit. We now have a `PopUpHandler.tscn` which then spawn `PopUp.tscn` since I was running into race conditions when I was trying to show Mana and Damage.

Because of this I have refactored the code a lot.

We now have one function that is connected with incoming signals `_play_anim` that now provides a `signal_type` based on the signal it receives.

This way we don't have a bunch of duplicated code.

`_play_anim` then spawns the old `PopUp.tscn` with the correct information and the `AnimationPlayer` now `queue_free`s when it's done since we don't need to reuse the same node.

godot/combat/interface/CombatInterface.gd
```diff
@@ -7,9 +7,11 @@ const CircularMenu = preload("res://combat/interface/circular_menu/CircularMenu.
 
 onready var lifebar_builder = $BattlersBarsBuilder
 onready var select_arrow = $SelectArrow
+onready var popup = $PopUpHandler
```
NathanLovato: Avoid "handler" in names. It makes it too easy to mess up your code with object oriented design. A handler could do many things from creating to animating the labels, which a single class shouldn't be responsible for. Spawner or builder, on the other hand, makes it clearer what this node should not do.

godot/combat/interface/CombatInterface.gd
```diff
@@ -7,9 +7,11 @@ const CircularMenu = preload("res://combat/interface/circular_menu/CircularMenu.
...
 
 func initialize(battlers : Array):
 	lifebar_builder.initialize(battlers)
+	popup.initialize(battlers)
```
NathanLovato: Godot already uses popup for popup windows. I'd prefer something like damage_label

NathanLovato:
Thanks for your work! I'll merge, rename a few things, and share the new commit back with you to show you how I'd change these.

PR #91, opened 2018-12-04
Fix #88 - missing skills now decrease mana properly

### LINKED ISSUE #88: SkillAction: If the skill misses, the battler doesn't lose mana

GresilleSiffle:


NathanLovato:
Thank you very much!

PR #100, opened 2018-12-04
Characters dialogue database and portraits added

### LINKED ISSUE #92: Dialogue: add characters database and portraits
For now, we can display the name of the character in a conversation, but is not tied to some reusable character data

To add portraits in multiple expressions per character, we need some kind of database. 
The idea is to have a Resource per character that contains all the information the interface needs to display it: a display name, maybe some background about the character, e.g. to display in the journal, and maybe a collection of textures to display to show the characters different expressions during dialogue.

Then, using the character's name/identifier in a conversation's JSON file should display the corresponding images on screen.

It be nice to display a full body character portrait as part of this task (reuse the battler images for Robi and Godette for now):

image

guilhermehto:
I opted to serve the textures with a function in the database that handles the fetching of the desired texture. It looked more natural and less error-prone than to add this logic to the Dialogue.gd since all it does is returning the Dictionary with the dialogue phrases, the actual building of the interface happens in DialogueBox. There, I fetch the textures by reaching to the database. 

Closes #92

### DIFF
--- a/godot/dialogue/characters/RobiDialogue.tres
+++ b/godot/dialogue/characters/RobiDialogue.tres
@@ -1,0 +1,14 @@
+[gd_resource type="Resource" load_steps=3 format=2]
+
+[ext_resource path="res://assets/sprites/battlers/robi.png" type="Texture" id=1]
+[ext_resource path="res://dialogue/characters/DialogueCharacter.gd" type="Script" id=2]
+
+[resource]
+
+script = ExtResource( 2 )
+display_name = "Robi"
+character_background = "Bingo bango bongo bish bash bosh"
+expression_textures = {
+"happy": ExtResource( 1 )
+}
+
--- a/godot/local_map/DialogueBox.gd
+++ b/godot/local_map/DialogueBox.gd
@@ -2,11 +2,13 @@ extends Control
 
 onready var dialogue_player : DialoguePlayer = get_node("DialoguePlayer")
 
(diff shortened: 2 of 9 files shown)


godot/dialogue/characters/RobiDialogue.tres
```diff
@@ -0,0 +1,14 @@
+[gd_resource type="Resource" load_steps=3 format=2]
+
+[ext_resource path="res://assets/sprites/battlers/robi.png" type="Texture" id=1]
+[ext_resource path="res://dialogue/characters/DialogueCharacter.gd" type="Script" id=2]
+
+[resource]
+
+script = ExtResource( 2 )
+display_name = "Robi"
+character_background = "Bingo bango bongo bish bash bosh"
```
NathanLovato: 😅

godot/local_map/DialogueBox.gd
```diff
@@ -2,11 +2,13 @@ extends Control
 
 onready var dialogue_player : DialoguePlayer = get_node("DialoguePlayer")
 
-onready var name_label : Label = get_node("Panel/Colums/Name")
-onready var text_label : Label = get_node("Panel/Colums/Text")
+onready var name_label : = get_node("Panel/Colums/Name") as Label
```
NathanLovato: Why have the : at all if you remove the type hint?

NathanLovato:
You could've built the database/dictionary from files on the disk instead of hard-coding it in a dictionary. Gonna add that now

PR #102, opened 2018-12-05
Refactor battlers to make it easier to edit and extend them

nhydock:
- Makes use of scene inheritence like it used to, making it a bit easier to actually see how parts plug into the results.  Battlers and their stats resources are still duplicated so that it's easier to reset combat.
- Combat Actions have been altered so they are self aware of if they can be used, which simplifies the API a bit.  
- Moved some calculations out of the Battlers and into the Combat Actions where suitable to help encapsulate the logic.
- Fixed a bug with life bars where if your max hp is higher than 100, the bar wouldn't be properly filled all the way.
- You can now unlock skills on level up
- Levels are defined as an experience growth lookup table
- Party Members now have a battler attached to them

The biggest drawback to using the node inheritance approach instead of abstracting with purely resources and a template container is that it's not immediately apparent which sections should be overridden without seeing another instance for example.

One of the most visible advantages is now you are able to properly visualize battlers when arranging formations.  Code has also been cleaned up/simplified a bit.


NathanLovato:
Yup it's looking good to me!

NathanLovato:
Great, even 😄

PR #104, opened 2018-12-05
Removed events from grid in local map

### LINKED ISSUE #50: Remove Grid being responsible for handling events
Currently the [`Grid.gd `](https://github.com/GDquest/godot-turn-based-rpg/blob/master/godot/local_map/grid/grid.gd) is responsible for handling encounters and objects interactions, this shouldn't be its responsibility, it should only handle movement.

What I think should be done is:

- It checks it a given cell is empty or not
- If empty it return the position
- Otherwise it emits a signal passing the object blocking the movement

In the [`LocalMap`](https://github.com/GDquest/godot-turn-based-rpg/blob/master/godot/local_map/LocalMap.gd) we handle these events by telling if there is any encounter, an object to be loot, an NPC interaction, a Dialogue to start.

So this whole [`block of code on Grid.gd`](https://github.com/GDquest/godot-turn-based-rpg/blob/master/godot/local_map/grid/grid.gd#L22-L36) will become something like:
 
```
  var cell_target_type = get_cellv(cell_target)
  if not cell_target_type == CELL_TYPES.EMPTY:
    var cell_pawn = get_cell_pawn(cell_target)
    emit_signal("cell_occupied", cell_pawn)
    return
  return update_pawn_position(pawn, cell_start, cell_target)
```

Then on `LocalMap` we handle the event based on the [`Pawn.type`](https://github.com/GDquest/godot-turn-based-rpg/blob/master/godot/local_map/pawns/pawn.gd#L4), in a method which the `Grid.cell_occupied` **signal** is connect to

guilhermehto:
Also created boiler plates for pawns that have some type of interaction with the player. They work with both raycasts (in case we want something like pokemon) and area2ds for something that should only trigger on contact. 

Closes #50

### DIFF
--- a/godot/local_map/LocalMap.gd
+++ b/godot/local_map/LocalMap.gd
@@ -3,11 +3,12 @@ extends Node
 signal enemies_encountered(formation)
 signal dialogue(dialogue)
 
-func _ready():
-	connect("dialogue", $MapInterface/Dialogue, "initialize")
-
-func _on_Grid_enemies_encountered(formation) -> void:
-	emit_signal("enemies_encountered", formation)
-
 func spawn_party(party) -> void:
 	$Grid/Pawns.spawn_party(party, $Grid.map_to_world(Vector2(2,2)))
+
+func _on_Grid_interaction_happened(type, arg):
+	match type:
+		InteractablePawn.InteractionType.DIALOGUE:
+			$MapInterface/DialogueBox.initialize(arg.load())
+		InteractablePawn.InteractionType.COMBAT:
+			emit_signal("enemies_encountered", arg.instance())
--- a/godot/local_map/grid/Grid.gd
+++ b/godot/local_map/grid/Grid.gd
@@ -1,13 +1,15 @@
 extends TileMap
 
-signal enemies_encountered(formation)
+signal interaction_happened(type, arg)
 
 enum CELL_TYPES { EMPTY = -1, ACTOR, OBSTACLE, OBJECT }
 
-onready var pawns = get_node("Pawns")
+onready var pawns = $Pawns
 
 func _ready():
 	for child in pawns.get_children():
+		if child is InteractablePawn:
+			child.connect("interacted", self, "_on_pawn_interacted")
 		if child is PawnFollower:
 			continue
 		child.position = request_move(child, Vector2(0, 0))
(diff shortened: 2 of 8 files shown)


nhydock:
Instead of matching based on some kind of strict type, it could be more useful for NPCs to have what they need injected into them and their interact functions would fully control how an interaction works.  You could cut down on the number of needed signal types and make the NPCs more powerful that way too.

godot/local_map/LocalMap.gd
```diff
@@ -3,11 +3,12 @@ extends Node
...
-	emit_signal("enemies_encountered", formation)
-
 func spawn_party(party) -> void:
-	$Grid/Pawns.spawn_party(party, $Grid.map_to_world(Vector2(2,2)))
\ No newline at end of file
+	$Grid/Pawns.spawn_party(party, $Grid.map_to_world(Vector2(2,2)))
+
+func _on_Grid_interaction_happened(type, arg):
+	match type:
+		InteractablePawn.InteractionType.DIALOGUE:
```
NathanLovato: InteractivePawn.Interaction.* ?

godot/local_map/LocalMap.gd
```diff
@@ -3,11 +3,12 @@ extends Node
...
-
-func _on_Grid_enemies_encountered(formation) -> void:
-	emit_signal("enemies_encountered", formation)
-
 func spawn_party(party) -> void:
-	$Grid/Pawns.spawn_party(party, $Grid.map_to_world(Vector2(2,2)))
\ No newline at end of file
+	$Grid/Pawns.spawn_party(party, $Grid.map_to_world(Vector2(2,2)))
+
+func _on_Grid_interaction_happened(type, arg):
```
NathanLovato: You can find a more descriptive name than `arg`. `data` ? As in interaction type, data

godot/local_map/grid/Grid.gd
```diff
@@ -1,13 +1,15 @@
 extends TileMap
 
-signal enemies_encountered(formation)
+signal interaction_happened(type, arg)
```
NathanLovato: The whole point of this task is to remove the grid's responsibility in interactions, to have the pawns interact with one another instead, so this doesn't fix the issue: https://github.com/GDquest/godot-turn-based-rpg/issues/50

godot/local_map/grid/Grid.gd
```diff
@@ -1,13 +1,15 @@
...
 
 enum CELL_TYPES { EMPTY = -1, ACTOR, OBSTACLE, OBJECT }
 
-onready var pawns = get_node("Pawns")
+onready var pawns = $Pawns
 
 func _ready():
 	for child in pawns.get_children():
+		if child is InteractablePawn:
```
NathanLovato: You should rename interactable -> interactive throughout the PR, interactable isn't an english word

godot/local_map/pawns/interactables/InteractablePawn.gd
```diff
@@ -0,0 +1,56 @@
+extends "res://local_map/pawns/Pawn.gd"
+class_name InteractablePawn
```
NathanLovato: InteractivePawn (don't forget to rename the file)

godot/local_map/pawns/interactables/InteractablePawn.gd
```diff
@@ -0,0 +1,56 @@
...
+export var sight_distance = 50
+export var facing = {
+	"up": true,
+	"left": true,
+	"right": true,
+	"down": true
+}
+
+var is_interacting : bool = false
+var interaction_arg
```
NathanLovato: It's not scalable, it'll soon need a rewrite: what if we want to read a dialogue, then start an encounter? If you need to store different types of data in the same variable and you can't assign a type to it, you may want to use an object. I think we should have some Interaction or MapAction class, similar to combat action, so we can have a list of commands on a pawn. It only takes a few extra minutes of work and it's scalable.

godot/local_map/pawns/interactives/actions/DialogueAction.gd
```diff
@@ -0,0 +1,13 @@
+extends MapAction
+class_name DialogueAction
+
+
+export var dialogue_box_path : NodePath
```
NathanLovato: Node paths are a bad idea for nodes that can be anywhere in the tree. 
And getting ancestors is a big big no-no in object oriented programming. You want to use groups and signals instead. 
I'll merge and refactor this time, it's simpler for me, but keep this in mind in general: never let children control their parent/ancestor. Unless you were to use components, but Godot doesn't support them well at all.

PR #117, opened 2018-12-10
Visual destination point added to pawn leader when touch mode is active

### LINKED ISSUE #114: Indicate where a character is moving in touch mode
Use a sprite, maybe an animated target or something. Should be part of the PawnLeader or on the UI?

guilhermehto:
I opted to add it to the leader, as a sprite. 
Closes #114


godot/local_map/pawns/PawnFollower.gd
```diff
@@ -2,6 +2,9 @@ extends PawnActor
 
 class_name PawnFollower
 
+func _ready() -> void:
```
NathanLovato: It should only be on the leader's scene. if you have to load and instantly free a resource at the start of the game without using it, you might as well not have it in the first place.

godot/local_map/pawns/PawnActor.gd
```diff
@@ -5,6 +5,8 @@ class_name PawnActor
 var game_board
 signal moved(last_position, current_position)
 
+onready var destination_point : = $DestinationPoint as Sprite
```
NathanLovato: Should be in PawnLeader as it's the only one that moves to destination

PR #119, opened 2018-12-11
InteractivePawn triggers actions with player input

### LINKED ISSUE #115: InteractivePawn should trigger actions when the player presses a key or touches it
They shouldn't have a "facing" dictionary either with up to 4 active directions at a time. To interact with an NPC, the player should have to:

- tap on it in touch mode
- press an action key on the keyboard or the controller

And automatic trigger upon walking in an area around the pawn should work differently: it should have some trigger area or something.

guilhermehto:
By getting near the pawn, a small balloon will show, allowing the player to press `ui_accept` to start the interaction. They can be started automatically by setting `auto_start_interaction` to `true`. Works with both Area2D triggers and the raycasts.

For the touch, it was already working as far as I could tell.

Closes #115


NathanLovato:
Touch works/worked but the UX is off: touching an NPC doesn't do anything: you've got to touch the cell next to it to move there

godot/local_map/pawns/InteractivePawn.gd
```diff
@@ -27,23 +27,37 @@ func _ready() -> void:
 		enable_process += int(raycast.enabled)
 	set_physics_process(enable_process > 0)
 
+func _process(delta : float) -> void:
```
NathanLovato: Why handle a one-time input in process?

PR #122, opened 2018-12-12
Small tweaks on Grid and LocalMap

salvob41:
Delete reference to map_size in localmap.tscn
instantiate class pathfinder directly with the class_name instead that preload gd file


NathanLovato:
Thanks!

PR #129, opened 2018-12-14
Fix #128, play the death animation correctly when a battler dies

### LINKED ISSUE #128: The 'death' animation is not playing when a battler dies
When a battler's health reaches 0, its `_on_health_depleted()` method is triggered but the death animation doesn't start.

GresilleSiffle:
Close #128


NathanLovato:
Thanks!

PR #136, opened 2018-12-18
Close #135, .gitignore now ignores all .import files

### LINKED ISSUE #135: The .gitignore file doesn't ignore all .import files
Currently not all .import files are ignored, even if the `.import` string is present in our `.gitignore`. As a result, all .import files present in the combat interface are always displayed as untracked by Git.

I think we should ignore `*.import` instead of just `.import` ([see the syntax here](https://git-scm.com/docs/gitignore)).

Gonna make a pull request.

GresilleSiffle:


NathanLovato:
See the replies on the issue, that's a no-no

PR #141, opened 2018-12-20
Fix clipping when the `highlight` animation of a `CombatPortrait` sta…

### LINKED ISSUE #140: UI: CombatPortrait: clipping when the highlight animation starts
There is a little clipping when the `highlight` animation of a `CombatPortrait ` starts.

I think this animation shouldn't modify the `modulate` property of the `CombatPortrait `.

GresilleSiffle:
…rts (close #140).

Now, the only animation that modifies the modulate property is `disable`.


NathanLovato:
Thanks for maintaining your patch :)

PR #143, opened 2018-12-21
Fix a bug that causes skills to be learned at the wrong level.

### LINKED ISSUE #142: Skills aren't learned at the correct level
So, for example, Godette's "Lollislash" in the inspector is set to be learned at level 1. However, when you run the game, Godette will not learn the skill until reaching level 3. Basically there's a 2 level difference

This has two different causes.

The first cause is simple. The script that determines whether a skill is unlocked or not, [here](https://github.com/GDquest/godot-turn-based-rpg/blob/98fa3c6b2b8c79d40d996efe9a19770d9737fcdf/godot/combat/battlers/skills/Skills.tscn#L22), should be using `<=` instead of `<`. The `<` unlocks the skill one level later than is intended.

The second cause... I'm not sure what to do with yet. But basically, in the script that displays the rewards, [here](https://github.com/GDquest/godot-turn-based-rpg/blob/98fa3c6b2b8c79d40d996efe9a19770d9737fcdf/godot/combat/Rewards.gd#L67), it displays the level as being one higher than it is internally. I am assuming this is because internally, characters start at level 0? Not sure how to fix this. Ideally the internal level would match the level we want to show to the player. I mean, a party member's level is determined [here](https://github.com/GDquest/godot-turn-based-rpg/blob/98fa3c6b2b8c79d40d996efe9a19770d9737fcdf/godot/combat/battlers/stats/GrowthStats.gd#L14), obviously, but I'm not sure yet whether or not changing things there will cause some sort of ripple effect for the rest of the game. You guys know what this affects, right?

I suppose the absolute simplest solution would be to make characters start with enough exp to be level 1 already. Then the displayed level could remain identical to the internal level.

If this approach is fine I've got a PR ready to go.

Dharengo:
Fixes #142

### DIFF
--- a/godot/combat/Rewards.gd
+++ b/godot/combat/Rewards.gd
@@ -64,7 +64,7 @@ func on_battle_completed():
 	$Panel/Label.text = "EXP Earned %d" % experience_earned
 	yield(get_tree().create_timer(2.0), "timeout")
 	for battler in leveled_up:
-		$Panel/Label.text = "%s Leveled Up to %d" % [battler.name, battler.stats.level + 1]
+		$Panel/Label.text = "%s Leveled Up to %d" % [battler.name, battler.stats.level]
 		yield(get_tree().create_timer(2.0), "timeout")
 	for drop in drops:
 		$Panel/Label.text = "Found %s %s(s)" % [drop.amount, drop.item.name]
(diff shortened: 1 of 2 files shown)


NathanLovato:
Thanks :)

PR #153, opened 2018-12-27
Basic quest system added

### LINKED ISSUE #150: Quest system
This issue's objective is to create a system that allows:
- [x] Creation of quests by designers 
- [x] Quest assignment to the player
- [x] Finishing quests

They should  as of now, support two types of objectives for them:
- Slaying monsters
- Talking with NPC's

There can be N number of objectives per quests.

guilhermehto:
Quests can be created by inheriting from `Quest.tscn`. The objectives are defined by adding `QuestObjective` children to it. There are currently two supported objectives: Interactions and Slaying monsters. 

- Interactions get completed when an interaction is finished with the defined pawn. I thought of allowing the user to define which action should complete the objective, but as we only have two main types of interactions (combat and dialogue) I decided to go with the simpler route. No need to future proof it now I believe.
- Slaying monsters is what the name suggests, when a combat is started, `died`, a new signal, is connected to all the slaying objectives in the quests. When the signal gets emmited, they check if that's the correct enemy and decrease the amount left to kill.

The `QuestSystem` is responsible for "managing" the quests. It's through it that they get added/removed from the active quest list. It was added to the `Game` scene and is passed to the `LocalMap` script. This way we can later give player quests based on events, i.e. the player entered a new area.

Another `MapAction` was created: `GiveQuestAction`. This action gives a quest to the player when interacted with. The new pawn created on the map uses it to give a `TestQuest` to the player. This quest has two objectives: kill 3 porcupines and speak with a newly created NPC. I'm not yet sure but we might have to extend the dialogue system: select which dialogue to display based on the player's active quests. 

Open to feedback on what can be improved in order to close #150 😃


godot/quest_system/QuestSystem.gd
```diff
@@ -0,0 +1,28 @@
+extends Node
+class_name QuestSystem
+
+var active_quests = []
+
+func add_quest(new_quest) -> void:
+	if has_quest(new_quest):
```
NathanLovato: You shouldn't have checks that just make a function do nothing if a user calls it in the wrong setting. We should never add the same Quest twice from the code!

godot/quest_system/QuestSystem.gd
```diff
@@ -0,0 +1,28 @@
...
+		return
+	add_child(new_quest)
```
NathanLovato: What is the difference between having the quest as a child node or having it in the array of active quests? How about having a Boolean value on the quest itself to tell whether it's active or not?

guilhermehto: We need the quests in the tree since they access it to connect to signals. I added a boolean to the quests to check if they are active or not and renamed the array to only `quests`. You could use `get_children()` but IMO the code is cleaner to read with the array.

guilhermehto:
@NathanLovato should be okay now

NathanLovato:
All good for now, merging!

PR #155, opened 2018-12-28
set initial animation for PawnAnim

ebuchmann:
This goes back to https://github.com/GDquest/godot-turn-based-rpg/issues/146 which was closed. There is still a odd jump when you first walk, which was due to `anim.get_current_animation_length()` getting 0 for the first step, and then 0.25 after that because an animation was set. This should fix it so it always gets the value on the first move.


NathanLovato:
Thank you very much 🙂

PR #157, opened 2018-12-28
Interaction only starts when clicking on top of InteractivePawns

### LINKED ISSUE #156: When next to an InteractivePawn, clicking anywhere on the screen starts the interaction
The interaction should only start when clicking on/touching the NPC itself

guilhermehto:
Since left mouse click event is on `ui_accept`'s action list, this would trigger the interaction. I created a check to see if the mouse is close to the pawn when the event occurs. The distance can be tweaked to make it easier to click on top of the pawn if the game is running on a mobile device, for instance. Tested with both mouse and touch screen.

Closes #156

### DIFF
--- a/godot/local_map/pawns/PawnInteractive.gd
+++ b/godot/local_map/pawns/PawnInteractive.gd
@@ -11,6 +11,8 @@ class_name PawnInteractive
 onready var raycasts : = $Raycasts as Node2D
 onready var dialogue_balloon : = $DialogueBalloon as Sprite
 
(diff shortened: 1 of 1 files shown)


NathanLovato:
Fixed manually in 90abed7b4331e1cb579d71745e4d6dcd7c4a90cc
You shouldn't use a radius: we have areas on the NPCs so that you can interact even if you're not next to the NPC, this should work with touch as well. Also no need to check for touch events as we emulate touch from mouse

Issue #152, opened 2018-12-27
UI: Quest journal


guilhermehto:
Should display the active quests and their progress

PR #162, opened 2019-01-02
Quest journal added

guilhermehto:
Here's a screenshot of how the journal currently looks like:

image

When you click the button on the bottom right corner, the menu slides in/out of the screen and the player is able to see the quests and their objectives. The items will get updated from signals sent from the `QuestSystem`.

I haven't found a way to access the tree items without having to use that `while true` together with `item.get_next()`. If anyone has worked with trees before, feel free to enlighten me on the subject and suggest changes.

Feedback on the overall UX of the menu is appreciated as well. 

The code is using mock data at the moment. I'll wait for #161 to get merged before hooking everything up.

This PR works towards closing #152


guilhermehto:
image

Now both the description and rewards are shown for quests. Objectives get their text updated (for progressive objectives i.e. slaying X monster), and their status. 

Code from `GUI.gd` has been removed and replaced with an initialization function as requested.

Scripts have been cleaned up a bit, and some improvements were made.

I decided to separate the notification system (to let the player know what rewards he got) from this issue, it would be too much work for only one PR, things would get mixed up, and the notification system can also be used for different things, thus I'll make it a bit more extensive than just a system to display acquired items. 

Should close #152 

@NathanLovato pinging you so you can see the changes.

godot/local_map/pawns/PawnQuestGiver.gd
```diff
@@ -0,0 +1,24 @@
...
+func _on_quest_given(quest : Quest) -> void:
+	quest.connect("quest_finished", self, "_on_quest_finished")
+	quest.connect("quest_delivered", self, "_on_quest_delivered")
+	quest_bubble.texture = QUEST_ACTIVE_TEXTURE
+
+func _on_quest_finished(quest : Quest) -> void:
+	if quest.has_to_be_delivered:
+		for action in actions.get_children():
+			if action is DeliverQuestAction:
+				quest_bubble.texture = QUEST_FINISHED_TEXTURE
```
NathanLovato: In general avoid going too deep with code blocks. No big deal here but it's part of the python (and gdscript) styleguide:
```
func _on_quest_finished(quest : Quest) -> void:
	if not quest.has_to_be_delivered: # or assert quest.has_to_be_delivered, the signal shouldn't be triggered for a quest that isn't finished, right?
		return
	for action in actions.get_children():
		if not action is DeliverQuestAction:
			continue
		quest_bubble.texture = QUEST_FINISHED_TEXTURE
		break
```

godot/local_map/pawns/PawnQuestReceiver.gd
```diff
@@ -0,0 +1,18 @@
+extends "res://local_map/pawns/QuestPawn.gd"
```
NathanLovato: Should be PawnQuest

Issue #89, opened 2018-12-04
SkillAction: When the skill misses, show a "miss" animation/label pop up


NathanLovato:

godofgrunts:
This probably could be combined with a status effect label if you wanted to use words for status effects (as opposed to sprites). But if you wanted to separate them (or maybe just have both) I don't think it adds that much complexity.

Probably should rename and `PopUpLabel` if we're going to separate them.

Not sure how you would want to go about that. Something like:

`PopUpLabelInt.tscn`
`PopUpLabelText.tscn`
`PopUpSprite.tscn`

And in `PopUpBuilder.gd` we should change `func spawn_label` to something like `func spawn_label_int` and make two more that are `spawn_label_text` and `spawn_sprite`.

NathanLovato:
It's a little abstract to me right now, I'm not sure what you would be trying to achieve with this setup. 
If you want to reuse the animations as resources between the nodes, they can't have different names.

Right now the element that comes up over the character is an animated control node. 
You could have anything as a child of it, use aggregation instead of separate base scenes to animate different elements popping up on top of a character.

Regarding localization, let's at least wait until we have the base game's design in place and words or sentences to translate. It's not a problem to add it later.

henriiquecampos:
@godofgrunts I think you can leave with just a single animation and just have a child, hidden, for each scenario, so a child for Miss, another for Stun, another for Damage, another for Healing, Mana...and show/hide them depending on what will be displayed. Since the animation itself will be on the parent, all of its children will be animated as well.

PR #184, opened 2019-01-14
Added Popup for missed skills

godofgrunts:
This should close issue #89

### DIFF
--- a/godot/combat/battlers/actions/SkillAction.gd
+++ b/godot/combat/battlers/actions/SkillAction.gd
@@ -22,7 +22,7 @@ func execute(targets):
 		for target in targets:
 			target.take_damage(hit)
 	else:
-		# miss
+		skill.emit_signal("attack_status", "Miss!")
 		pass
 		
 	yield(actor.get_tree().create_timer(1.0), "timeout")
--- a/godot/combat/battlers/actions/SkillAction.gd
+++ b/godot/combat/battlers/actions/SkillAction.gd
@@ -22,7 +22,7 @@ func execute(targets):
 		for target in targets:
 			target.take_damage(hit)
 	else:
-		skill.emit_signal("attack_status", "Miss!")
+		skill.emit_signal("missed", "Miss!")
 		pass
 		
 	yield(actor.get_tree().create_timer(1.0), "timeout")
--- a/godot/combat/battlers/skills/Skill.gd
+++ b/godot/combat/battlers/skills/Skill.gd
@@ -2,7 +2,7 @@ extends Resource
 
 class_name Skill
 
-signal attack_status(text)
+signal missed(text)
 
 export var name : String = "Skill"
 export var description : String = ""
--- a/godot/combat/battlers/skills/Skill.gd
+++ b/godot/combat/battlers/skills/Skill.gd
@@ -2,6 +2,8 @@ extends Resource
 
 class_name Skill
 
+signal attack_status(text)
+
 export var name : String = "Skill"
 export var description : String = ""
 export var icon : Texture = load("res://icon.png")
--- a/godot/combat/interface/PopupLabel.gd
+++ b/godot/combat/interface/PopupLabel.gd
@@ -5,23 +5,20 @@ var _animation_name : String
 
 export var offset : Vector2 = Vector2(0.0, -40.0)
 
-func initialize_damage(battler : Battler, type : String, difference : int):
+func initialize(battler : Battler, type : String, message : String):
 	"""
-	@type: either health or mana. Determines the animation the label will use
+	@type: either health, mana, missed. Determines the animation the label will use
 	"""
-	assert type in ["health", "mana"]
+	assert type in ["health", "mana", "missed"]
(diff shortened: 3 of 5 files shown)


godot/combat/battlers/actions/SkillAction.gd
```diff
@@ -22,7 +22,7 @@ func execute(targets):
 		for target in targets:
 			target.take_damage(hit)
 	else:
-		# miss
+		skill.emit_signal("attack_status", "Miss!")
```
NathanLovato: Signals should be past tense and represent an event. You know that the skill emits it, and a skill may not be an attack. The signal should be `missed`.

godot/combat/battlers/skills/Skill.gd
```diff
@@ -2,6 +2,8 @@ extends Resource
 
 class_name Skill
 
+signal attack_status(text)
```
NathanLovato: `signal missed`

godot/combat/interface/PopupLabel.gd
```diff
@@ -5,7 +5,7 @@ var _animation_name : String
 
 export var offset : Vector2 = Vector2(0.0, -40.0)
 
-func initialize(battler : Battler, type : String, difference : int):
+func initialize_damage(battler : Battler, type : String, difference : int):
```
NathanLovato: There's an issue here. You shouldn't make different initialize methods. If you feel like you need to break down initialize into two, encapsulate that in the script, but don't add another method to the public interface that you will call from the outside - every time you do so, you add a dependency/coupling with this class.

godot/combat/interface/PopupLabelBuilder.gd
```diff
@@ -4,12 +4,20 @@ const label_scene = preload("PopupLabel.tscn")
...
 	for battler in battlers:
-		battler.stats.connect("health_changed", self, "spawn_label", [battler, "health"])
-		battler.stats.connect("mana_changed", self, "spawn_label", [battler, "mana"])
+		battler.stats.connect("health_changed", self, "spawn_label_number", [battler, "health"])
+		battler.stats.connect("mana_changed", self, "spawn_label_number", [battler, "mana"])
+		for skill in battler.skills.get_children():
+			skill.connect("attack_status", self, "spawn_label_text", [battler, "miss"])
 
-func spawn_label(new_value : int, old_value : int, battler : Battler, signal_type : String) -> void:
+func spawn_label_number(new_value : int, old_value : int, battler : Battler, signal_type : String) -> void:
```
NathanLovato: From the comment about 2 initialize methods above: numbers and text can both be text. The issue here is you should give the label the text to display, on which battler, and nothing else.

godofgrunts: My issue here is that `health_changed` and `mana_changed` send the new value and old value which I then subtract to get the final number to display.

I don't know what systems already depend on that function emitting those numbers and I'm not sure how I could incorporate that into the one initialize function.

godot/combat/battlers/actions/SkillAction.gd
godofgrunts: Done

godot/combat/battlers/skills/Skill.gd
godofgrunts: Done

godot/combat/interface/PopupLabelBuilder.gd
NathanLovato: You should get the values from the signals in a given node, likely the Builder, prepare the text you want to display: whether it's a number or a message. Then just create the animated label with that text. Right now you're duplicating this responsibility between the builder and the label node, making your code more fragile than it needs to be: if we need to make a change to any signal upstream it will break two classes.

NathanLovato: If you don't see how to refactor this tell me, I'll create a PR to your fork directly.

godot/combat/interface/PopupLabelBuilder.gd
```diff
@@ -4,20 +4,17 @@ const label_scene = preload("PopupLabel.tscn")
...
+		battler.stats.connect("mana_changed", self, "convert_numbers", [battler, "mana"])
 		for skill in battler.skills.get_children():
-			skill.connect("attack_status", self, "spawn_label_text", [battler, "miss"])
+			skill.connect("missed", self, "spawn_label", [battler, "missed"])
 
-func spawn_label_number(new_value : int, old_value : int, battler : Battler, signal_type : String) -> void:
-	var difference = new_value - old_value
+func convert_numbers(new_value : int, old_value : int, battler : Battler, signal_type : String) -> void:
+	var difference = str(new_value - old_value)
+	spawn_label(difference, battler, signal_type)
```
NathanLovato: A method `convert_number` shouldn't call `spawn_label` -> make methods just return a value whenever possible. I'll merge and do a quick refactor pass now

NathanLovato:
There you go! https://github.com/GDquest/godot-open-rpg/commit/04c03d9a234571dbfb8a9671f6ab1e7cc5b15387

PR #191, opened 2019-06-27
Fix #190: Enabling Pixel Snap

### LINKED ISSUE #190: One gray pixel line appear on tiles, when the camara moves
I'm working on Ubuntu 16.04.06 64bits
I test it with the editor, and compiled version, and still the result is the same

image
image
image

MateoMiccino:
This PR fix #190 issue


NathanLovato:
Thanks!

PR #202, opened 2021-02-08
Use drop chance correctly

### LINKED ISSUE #201: Drop chance is ignored
**I'm submitting a...**

- [x] Bug report.
- [ ] Feature request.

## Bug report ##

**What is the current behavior?**

The drop chance is ignored

**What is the expected behavior?**

Whatever we get a drop is determined by its chance

**Tell us the steps to reproduce the bug, and if possible share a minimal demo of the problem.** 

Read the code at [Rewards.gd line 27](https://github.com/GDQuest/godot-open-rpg/blob/eac5d4454cac7d2f40d7b383852ff989c9d88183/godot/src/combat/Rewards.gd#L27), it currently says:

```
if drop.chance - randf() > drop.chance
```

So, we subtract from chance a positive number between 0 and 1 (see [randf](https://docs.godotengine.org/en/latest/classes/class_@gdscript.html#class-gdscript-method-randf)), and we check if that is greater than chance. I repeat, we subtract a positive, and check if we ended up with more.

There are two ways this expression can go: We get false, or we get a type error.

If we create a new project and use it to run the following code:

```
func _ready():
	check(0);
	check(1);
	check(-1);
	check(10);
	check(-10);
	check(0.5);
	check(-0.5);
	check(INF);
	check(-INF);
	check(NAN);
	check(-NAN);
	
func check(x):
	print(x - 1.0 > x);
	print(x - 0 > x);
```

The output will be `False` 22 times.

theraot:
Given that `randf` output is in the range [0, 1] inclusive, and we want `chance = 0` to mean never, and `chance = 1` to mean always, we need to worry about the edge cases.

We need the `>=` because when `randf() == 0` and `chance == 0` just using `>` would give us the wrong result.

We need to compare against 1 because when `randf() == 1` and `chance == 1` we would get the wrong reasult due to the `>=`.

This is the from I found easier to read.

**Please check if the PR fulfills these requirements:**

- [ ] The commit message follows our guidelines.
- For bug fixes and features:
    - [x] You tested the changes.
    - [ ] You updated the docs or changelog.

Related issue (if applicable): fixes #201

**Other information**

I had to create a DropItem script for testing the drop. It is not included in the pull request. I'm trying to fix, not add.

### DIFF
--- a/godot/src/combat/Rewards.gd
+++ b/godot/src/combat/Rewards.gd
@@ -24,7 +24,7 @@ func _add_reward(battler: Battler):
 	# Appends dictionaries with the form { 'item': Item.tres, 'amount': amount } of dropped items to the drops array.
 	experience_earned += battler.drops.experience
 	for drop in battler.drops.get_drops():
-		if drop.chance - randf() > drop.chance:
+		if drop.chance < 1 and randf() >= drop.chance:
 			continue
 		var amount: int = (
 			1


NathanLovato:
Thanks!

PR #212, opened 2023-06-08
Add Cutscenes: trigger-able sequences of events such as dialogues, treasure chests, & area transitions

food-please:
**-Please note this is currently a draft PR and will be updated when ready for review.-**
I'm anticipating two review cycles: one to verify preliminary event design (see comment at the relevant [issue](https://github.com/GDQuest/godot-open-rpg/issues/211) and another to verify the full change list implemented by this PR.

**Please check if the PR fulfills these requirements:**

- [x] The commit message follows our guidelines.:
- [x] You tested the changes.

Related issue: [211](https://github.com/GDQuest/godot-open-rpg/issues/211)

**What kind of change does this PR introduce?**
- Added different kinds of events, objects easily extended with Godot's node/script system that allow modifying the field game state. Examples include dialogue, adding an NPC, changing map tiles, etc. I'm planning on two specialized types of Event:
   - **Triggers**, which are objects placed directly on the gameboard that are "triggered" by a collision object (Gamepiece) moving onto them. Explicitly tracks which gamepiece triggered the event.
   - **Interactions**, which are Events attached directly to a Gamepiece that are activated by "interacting" (for example, press space or the A button) with the gamepiece from a neighbouring cell. Explicitly tracks the parent and interacting gamepieces.
- Added a simple MusicPlayer that makes it easier to play a given background track.
- Documentation: small updates & new features

** TODO: remaining features before PR ** (edited 2023-10-29)
- [x] JRPG-style interactions. E.g. "press space to interact" with object directly in front of the player
- [x] Add a handful of derived, standard interactions. Examples include doors (locked, with variable unlocking criteria), treasure chests, hidden paths, and dialogues.
- [x] Dialogue via Dialogic2
- [x] Optional "popups" of some sort to show player which objects are interactable
- [x] Temporary rudimentary "inventory" to demonstrate conditional interaction logic. Will be expanded in future updates.
- [x] Fix controllers/pathfinding to make more useful. Base controllers should be able to move gamepieces along a path with a simple function call. Also there are a few bugs that need to be ironed out.
- [x] Pause player/NPC movement and input while interactions are running
- [x] Music
- [x] Simple area (e.g. exterior to interior) & screen (e.g. fade) transitions
- [x] Demo interaction scene w/ clear **beginning** and **ending**. Probably includes blowing something up with a bomb.

~~- [] Popup icons (illustrate important NPCs/Events, etc.)~~
~~- [] MusicPlayer::crossfade()~~
~~- [] Dialogue via Dialogic2~~
~~- [] "Cinematic" events that pause input/controllers while they're executing (to prevent wandering off).~~
~~- [] After a review, full documentation and a demo scene incorporating more complex events.~~

**Does this PR introduce a breaking change?**
No, should be focused on adding functionality.


NathanLovato:
Thanks so much once more for your work! Regarding review, you mention needing a round on the events system code. Would you like it at this stage already or would you like to do some more work before we give you feedback?

NathanLovato:
Sure! Note taken, we'll wait for you to tell us when you'd like a review.

food-please:
Good day,

I think I have a little scene that should suitably demo events now. The demo scene itself is hacky; it is just meant to demo the event scripts. In particular I'm looking for feedback on the [event](https://github.com/food-please/godot4-open-rpg/tree/main/src/field/gameboard/events) objects, though feedback on everything else is welcome also.

The flow of the demo is as follows: click (or move up to and then press 'SPACE' to interact) the guy blocking the road and he'll move out of the way. Walk over the key to pick it up and then walk up to and interact with (via WASD and the SPACE bar) the locked fence. Once that's out of the way, you can move over to the exit.

What I'm trying to demonstrate is how events are nothing more than fancy physics objects that can and should be extended by scenario designers to perform more complex tasks. With this in mind, an event is designed to be Godot-y and GDQuest-y as it is nothing more than a scene that can be attached to Gamepieces, TileMaps, nothing at all, etc. The various events in the demo make use of Godot nodes to control the flow of the map.

There are two subclasses of events: interactions (which can be interacted with via SPACE or clicks/touches) and triggers (which are triggered by Gamepieces walking over).

Please let me know what you think. Once we've got this wrapped up I'll try to finish off the milestone with Dialogic 2 support. Also, I'll come up with a cleaner demo scene to show off events in a more polished fashion.

Thanks again.

NathanLovato:
Just so you know we are going to an industry event on Sunday so we are busy preparing for it but I took note of your progress and added the review to my to-do list. We will do it as soon as possible.

Xananax:
Very good work! I like the demo; I found it a bit difficult to understand the level of feedback required at this stage. Is the demo representative of the kind of interactions desired in the end result? If yes, it seems to me that your implementation could be simplified.

A few generic impressions and questions:

1. Naming is a bit confusing. I kept looking for how an `Event` replaces a `Signal`, and it took me a minute before realizing that was not it. By Event, you mean an EventObserver or an EventListener.  Something that, under the correct circumstances, will trigger an event. Similarly, what you call an `Interaction` is something that will trigger an interaction, etc. This is a bit confusing. The names we landed upon internally for those things are `Interactable`, `Trigger`, and `Pickable` . An event is something that "happens" when an Interactable triggers. "Interactable" is a bit of a mouthful, for certain, but it does get the meaning across.
2. Are we considering the case where a key to a door is in a different scene (and therefore the key has no access to the door)?
3. Are we considering the case where a key can open any door of a certain type, but then gets consumed (Zelda-like)?
4. Are we considering cases with multiple potential interactions, or will we always pick the one the player is turned towards? 
5. Why is "CinematicEvent" part of the Interactable? It seems like it should be part of an autoload or the board manager. I'm not clear on why calling `cinematic_mode_enabled` can't be something called from an interactable on the `GlobalFieldEvents`, when needed. I'm also not clear on the need for the intermediary `CinematicEventHelper` object. Is this object only for awaiting for all pieces to reach their destination? If yes, then I would propose it should be a generic "await_all_pieces" function attached to `GlobalFieldEvents`, or be renamed to something suggesting the function better than "helper".
6. What are interaction groups for? Do you see them serving a later purpose?
7. Why can Interactables be turned off and on? Do you envision a need for this?

I can follow up with more feedback depending on the answers, especially of 2, 3, and 4.

It seems to me that you can simplify the way you do the interactions and still check all the same goals you currently have (but maybe not if there are future goals not demonstrated). 
I've written a quick example here https://github.com/GDQuest/godot-open-rpg/tree/temp-branch-for-demonstration

You might want to look at this repository, written by a collaborator for a yet unreleased demo: https://github.com/Ryan-Mirch/Interactables-Demo, which tackles the same area of problems.

food-please:
> Is the demo representative of the kind of interactions desired in the end result? 

My apologies, I have probably caused some confusion. My intention is to create the groundwork needed for any "kind" of interaction that would be implemented down the road. I wanted to keep it as simple as possible, so most of the interaction code 

> A few generic impressions and questions:
> 
> 1. Naming is a bit confusing. 

Noted, I'll run with these. I'm not a huge fan of the alternatives but I can't think of anything better that doesn't overlap engine features/programming paradigms.

> 2. Are we considering the case where a key to a door is in a different scene (and therefore the key has no access to the door)?
> 3. Are we considering the case where a key can open any door of a certain type, but then gets consumed (Zelda-like)?

No, not yet. In my mind that is outside the scope of this demo. I think it's application specific and its implementation is left to the designer. The only point of the key/lock relationship was that I wanted to demonstrate how scene-based interactions can easily communicate and change one another's state.

> 4. Are we considering cases with multiple potential interactions, or will we always pick the one the player is turned towards?

In my mind it makes most sense to search for the top pickable interaction that the player is facing. I think that is expected behaviour with a gamepad/keyboard. Touch may be a little different, but that would require minor changes work correctly. That functionality should be included in this particular release.

> 5. Why is "CinematicEvent" part of the Interactable? It seems like it should be part of an autoload or the board manager. I'm not clear on why calling `cinematic_mode_enabled` can't be something called from an interactable on the `GlobalFieldEvents`, when needed. I'm also not clear on the need for the intermediary `CinematicEventHelper` object. Is this object only for awaiting for all pieces to reach their destination? If yes, then I would propose it should be a generic "await_all_pieces" function attached to `GlobalFieldEvents`, or be renamed to something suggesting the function better than "helper".
> 6. What are interaction groups for? Do you see them serving a later purpose?

The main point here is that it seemed good to me to have some interactions that "take control" of the application and some that keep happily on their way in the background. This may introduce complexity we don't want, since it implies concurrent interactions.  I would hesitate to make the "cinematic mode" capability global, as only interactions should be doing so, in my mind.

The point of the groups was that I had been thinking active interactions will need to be paused when combat starts and resumed when it finishes. We would need a way to grab active interactions and pause/resume them and groups seemed the easiest way to do so.

It may be a better idea to have an event queue and only allow one event to run at a time. Thoughts?

> 7. Why can Interactables be turned off and on? Do you envision a need for this?

I think this would be useful. Consider a designer wanting to drop a "spawn point" into a scene but it shouldn't activate until later in the game. You could either create the node at runtime and have it appear when you need it or you could drop it in during map design, disable it, and activate it when the time comes. I prefer the latter for a few reasons:

- I think it enables elegant design. In my mind adding nodes at runtime is more complex (error prone) than baking it into the structure. I don't think that the designer shouldn't have to keep track of what will appear down the road.
- The interaction is visible in-editor. In my mind this helps with blocking out a map.
- It makes re-using events easier. Just toggle them on/off rather than freeing/new-ing similar interactables. They certainly don't have to be reused, but think that I can think of interactions that would benefit from it.
- There may be drawbacks or other perks I haven't thought of. What are your thoughts on the matter?

> It seems to me that you can simplify the way you do the interactions...
> You might want to look at this repository, written by a collaborator for a yet unreleased demo: https://github.com/Ryan-Mirch/Interactables-Demo, which tackles the same area of problems.
> I can follow up with more feedback depending on the answers, especially of 2, 3, and 4.

Thanks. I'm going to revisit this ASAP, so I'll take a look through the other projects and, if you had any more features/ideas to simplify things I would love to incorporate them at that time.

Thanks for the review, this is what I was looking for.

Xananax:
> [...] My intention is to create the groundwork needed for any "kind" of interaction that would be implemented down the road. 

That doesn't really work well imho.

Warning: pontificating paragraph with Strong Opinions:registered: ahead.

There are two ways of making something modular:

The first is implementing provisions for hypothetical future requirements. The proper way of doing this is by defining an extremely solid API with enough flexibility to allow for novel usages, and enough discoverability so people can use it. In practice, this often means a few hundred lines of code for sorting items by arbitrary values can be replaced in your specific game with `items.sort_custom(func (a, b) -> a.weight < b.weight)`. There may be many occasions where the framework, no matter how flexible, does not allow exactly what you need even.

In other words, in this first option, you're accruing an enormous amount of complexity, maintenance burden, tech debt, code branching, bugs, innefficiency. Its very rarely worth it. It's worth it when you're either abstracting very complex code (physics engine), or offering new data structures that can be plugged into each other (a hashmap).

But often, it's better to consider the code itself as your API. Godot is already providing you with a very versatile framework. Giving a user `items.sort_custom(...` is many times easier for them to understand and change than a 200 LOC framework. It is objectively more flexible too.

So, the second way of making something modular is to make it strictly as inflexible as conceivable. Through simplicity, modularity is achieved; rewriting a line of idiomatic code is many times simpler than understanding a newish use-once-and-throw spec.

This was a long preamble (sorry) to say: I would much rather you define a narrow set of behaviors you actually do want to implement, and write the simplest code that achieves that. Users who want more can scavenge the pieces of code they want and write a new, as straightforward system, with just as little indirection.

> I'm not a huge fan of the alternatives but I can't think of anything better [...]

Same here! You're welcome to propose something.

> The main point here is that it seemed good to me to have some interactions that "take control" of the application [...]

Fine, that sounds fair. I agree with your take.

> The point of the groups was that I had been thinking active interactions will need to be paused when combat starts and resumed when it finishes. We would need a way to grab active interactions and pause/resume them and groups seemed the easiest way to do so.

I feel like this is premature. I suggest crossing that bridge when we get to it. My reasoning is:

1. We might never have combat while an interaction is happening (in most traditional games, those two things are in areas that don't intersect)
2. We might have combat as an interactable (even if randomly spawned). In this case, there may be a general need for waiting for interactables to finish, which may happen by registering to some global manager so no two events can overlap ("event queue", as you suggest).

I don't mind too much the groups, they aren't adding that much complexity; I just don't think they are necessarily the best way of tackling this (as of yet, hypothetical) problem.

Instead of an event queue, we might just register and deregister one event when it starts and ends, and if the player tries to trigger another, we display a "no" icon or play a bleep sound. Not all events would consult that registry (for example, pickups maybe don't need to and can continue running).

> Consider a designer wanting to drop a "spawn point" into a scene but it shouldn't activate until later in the game

Yup, good call. I'm down for enabling/disabling.

> the main difference in architecture at the moment is that the demo handles management/picking via autoload whereas the OpenRPG delegates that to the player controller.

From experience, having a global handler to delegate interactions to often ends up being a better option because there's often external, global information you'd want to know in order to know what to trigger. Examples:

- Overlapping Interactables (which one to pick?)
- Triggering the interactable the player clearly _meant_, because it's an important one, over the one they're turned towards, even if it'd be the correct action
- Needing to access other globals (eg, UI timer when switch is triggered, tutorial prompt "press X to interact")
- Needing to access some global state to know if an interaction is valid or not (are we in a cinematic mode?), and maybe act differently depending on that state (show a UI "skip" prompt)

I'm not suggesting it's the only solution or even appropriate for the current demo, but just to consider the option going forward.

> I had been thinking that the CinematicEventHelper would handle the actual implementation of cinematic event. Now that I think about it that's probably what you were thinking would become an autoload. Is that correct?

It really depends on what kind of cinematic event you're envisioning and how they are implemented tbh, so I will not expand on this before having a better idea of what you're aiming for.

----

In reference to my opening paragraph, I would suggest to determine a scenario for the MVP, in bullet point format. Something like:

1. Player picks a key
2. Player enters a fight (fight can be auto-resolved and implemented later)
3. Player opens a chest and gets a key
4. Player enters a house; cinematic event is triggered and we see two characters arguing
5. ...

Trying to nail all the basic interactions that are necessary. The interactions, dialogues, graphics, etc, can be implemented as simply as possible, but the list should be relatively comprehensive and demonstrate all the interactions.

We don't need to go ahead and implement all, or maybe any of them; but they will guide the decisions regarding how to slice this cake (who has the responsibility to do what; autoload vs instance; etc).

food-please:
I'd like to do two things here. I'll write a response and then write a second post to try to shape the MVP.

> That doesn't really work well imho...
> ...This was a long preamble (sorry) to say: I would much rather you define a narrow set of behaviors you actually do want to implement, and write the simplest code that achieves that. Users who want more can scavenge the pieces of code they want and write a new, as straightforward system, with just as little indirection.

Makes sense to me, I think my initial struggle is understanding what particular elements should be simpler (i.e. the whole system? The 'cinematic mode'? The reliance on a PlayerController?). I'm still trying to wrap my head around that slightly, so I'll do a rework based on your above writeup as well as the demo(s).

How ready is Ryan Mirch's interactable demo? If the GDQuest team is relatively happy with where it stands, I would consider porting the "inheritance" model to 2D and add any attribution required and requested. I'm perfectly happy to let someone else's code solve the issue.

> 1. We might never have combat while an interaction is happening (in most traditional games, those two things are in areas that don't intersect)

There are a couple of options. At this point I'm not sure what is best:

- End active interactions on combat start. This is probably the simplest. Interaction execution would need to know to break when starting combat. I think you would then want the combat object (whatever that looks like) to have an on_loss/on_victory script/interaction that picks up afterwards, if necessary.
- Pause currently active interactions on combat start. This is (much?) more complicated but allows for persistent, concurrent interactions.

> 2. We might have combat as an interactable (even if randomly spawned). In this case, there may be a general need for waiting for interactables to finish, which may happen by registering to some global manager so no two events can overlap ("event queue", as you suggest).

That's an interesting idea. I had originally been thinking that combat itself would be a scene and the designer could be a scene per enemy formation/background. I presume a combat interaction would just hold "combat data".

> Instead of an event queue, we might just register and deregister one event when it starts and ends, and if the player tries to trigger another, we display a "no" icon or play a bleep sound. Not all events would consult that registry (for example, pickups maybe don't need to and can continue running).

This is probably the answer for 'cinematic_mode'. I might rename that to just cutscene, as I think that's less wordy and more ubiquitous in the industry. At any rate, you only want one running at a time, though it is conceivable that two or more could activate at the same time and we would need to either queue each or drop all but the first. I'm still on the fence here.

>... but just to consider the option going forward.

> It really depends on what kind of cinematic event you're envisioning and how they are implemented tbh, so I will not expand on this before having a better idea of what you're aiming for.

The only criteria for a cutscene in my head is:
a) exterior stimuli (controllers, signals, etc.) are ignored
b) has a definitive start and endpoint.

> In reference to my opening paragraph, I would suggest to determine a scenario for the MVP, in bullet point format. 

I'll do this in the post below.

food-please:
Regarding the MVP, I'm not entirely sure about scope, but I would like to try to capture a handful of basic (polished!) features expected in a JRPG. I feel that some of these are typically overlooked in tutorials since they involve game state and persistence.

MVP rough outline:
- Brief intro (text, maybe pictures over black screen) to scenario
- Reveal screen, characters walk in, quick line of text introducing character to scenario.
- Player finds a chest with a key inside
- Player uses key in lock to travel through door.
- Player travels to the next area.
- Player finds a 'monster' blocking the exit. Has a brief conversation before entering a mock combat/victory state. 
- Player can now walk through path to exit.
- After player exits, run an event BEFORE screen fades to wrap up scenario.

I'm assuming that interactions will still be physics-based components as already designed. There are a few "interaction types" going on here:
- Interactable, run when clicking or pressing the interact key
- Trigger, run when a corresponding CollisionObject2D arrives at its cell.
- Script (of some sort) that can be run when entering or exiting an area, losing or winning combat, etc.
   - Probably should be able to call these arbitrarily, event from other interactions. As in scripted_event.run().
   - Probably should be able to be cutscene/exclusive, also.
   - Probably should be scenes in order to build the same way as Triggers/Interactables. Do NOT need to be derived from Area2D. Therefore, if these are all related. Triggers and Interactables could derive from here and have collision info added in their own scenes. The advantage to this is that any common functionality needed by interactions (e.g. change music, get player character, start combat, etc. etc.) could then be applied to the base class rather. 

I anticipate a few often-used subclasses:
- Lootable extends Interactable: treasure chests!
- Pickup extends Trigger: adds item to inventory and disappears.
- MapExit extends Trigger: fade out, move to destination area, probably have an on_exit script that can be run before the transition

Changing the map scene is a significant chunk of code. I don't see any reason to wait to implement it however.

Changing to and from the combat state is also pretty major. This isn't a bad time to add in the system structure, but it would be quickly auto-resolved.

As mentioned, we might not get to all of the features at the moment. I would anticipate two pull requests here: one with the interaction implementation and some simple interactions and one with changing maps/demo combat. I would like to wrap up the interaction implementation next. We can always discuss game state next.

Would you change any of this? Thanks so much for your time, I hope that I can get this in good shape without bothering you too much more.

Xananax:
I think it's a very good next step.

I am not bothered at all, it's a pleasure working with you, no worries and feel free to ping as many times as you like.

> How ready for use is Ryan Mirch's interactable demo? 

This demo is part of a new resource we're developing with recipes; the principle isn't to use the demos as they are (though it's certainly possible) but to get inspired and implement something more specific. That's why we provide different competing implementations for each mechanic.

To answer the question more directly, I've reviewed the code and I am personally satisfied with it, but it was written in a nebulous vague context that isn't specific to any game, so I'd invite you to be inspired, rather than use outright; maybe scavenge some parts.

At the moment, if there's no issue with centralizing the interactions in the player, then there's no issue, so I'd say to keep it as it is. I only showed this repo to compare and contrast details. Feel free to use any amount of it, including none.

> I presume a combat interaction would just hold "combat data".

It could remain a scene, yea, but we need some sort of trigger. I figure, some combat scenes (story ones) will be triggered by conversations or entering rooms anyway, so we'll need the `CombatTriggerer` interactable either way, even if random encounters are based on a timer/probability per step or something.

I personally also prefer visible encounters that can be avoided (which would be interactables too) but that really depends on the type of game you have in mind.

Since your scenario includes a visible, triggerable encounter, I think we're on the same general wavelengths.

> Script (of some sort)

I've never tried this pattern, but do consider that you can have an expression in a textfield and/or a "runscript" field in which you could directly write some GDScript.

<details>

<summary>(Quick Pseudocode example)</summary>

```gdscript
var _expression := Expression.new()
@export var expression_str := "":
	set(value):
		var error := _expression.parse(value, ["arg_a", "arg_b"])
		if error != OK:
			print(_expression.get_error_text())
			return
		expression_str = value

func exec_expression():
	var result = _expression.execute("foo", "bar")
	if not _expression.has_execute_failed():
    print(result)

# or with a script

var _script := GDScript.new()
@export var script_str := "":
	set(value):
    var lines := ["func run(arg_a, arg_b):", "  var result = null"]
	  lines.append_array(Array(value.split("\n")).map(
			func (line: String) -> String:
				return "  "+line
		))
    lines.append("  return result")
		_script.source_code = "\n".join(lines)

func exec_script():
  var instance = _script.new();
  if not instance.has_method("run"):
    pusherr("oh noes")
  var result = instance.run("foo", "bar")
  print(result)
```
</details>

@NathanLovato did you ever try something like this? I imagine it ends up being problematic at scale but maybe a lot of ad-hoc scripts gets also unyieldy.

At first sight I don't like this but it may be worth considering if you intend to have ad-hoc specific one-off triggers. I'm not sure if that is what you meant.

----

In general, I think your scenario and plans to implement changing scenes and combat sound and appropriate. Feel free to go ahead, and thank you for the excellent work!

NathanLovato:
> @NathanLovato did you ever try something like this?

No but I wouldn't attempt this either, letting users input a string in a field. Note that Expression only supports basic math operations, it's not related to GDScript.

For this project, I'd say just let people write a GDScript file for every event sequence. If something comes up a lot and is too tedious to repeat with GDScript, then I'd say you can have predefined nodes, scenes, or resources with a couple of exported variables.

food-please:
Regarding node naming, would Trigger (base class), ContactTrigger (triggered by movement), and InteractionTrigger (triggered by 'interact') make sense? The idea with "trigger" is that it is usually "triggered" via some kind of external stimulus.

Or perhaps something like Triggerable (base class), Contactable (movement), Interactable (interaction)?

Clear and unique enough? Too verbose? I wasn't sure if they should be unified (as in, KinematicBody -> KinematicBody2D). I'm probably overthinking this.

Xananax:
> Regarding node naming, would Trigger [...] or perhaps something like Triggerable [...]

I think both are fine yea, they seem to explain adequately the purpose for me.

food-please:
Another quick question, do you have any experience with [Dialogue Manager](https://github.com/nathanhoad/godot_dialogue_manager)? Do you have a preference between that and Dialogic?

The reason I ask is that Dialogic 2 is still in alpha and some of the core features are changing while I'm putting the demo together. It would be nice to have something stable.

The other issue I gather is that Dialogue Manager is simpler. From a quick playing around with the plugins I get the impression that it would be easier to move Dialogue Manager -> Dialogic 2 than the other way around (if Dialogue Manager was deemed inadequate someday down the road).

Xananax:
I don't personally have a very strong preference towards one or the other. 

I tend to think that if you want to diverge from natural/native systems to the framework, going Dialogic, which is expansive enough to be its own framework, is better than doing something half-half, that limits you somewhat for the benefit of a relatively small increase in readability. More on this in a (long and non-mandatory) explanation below. "Go big or go home", etc.

But I'm ok with either, it's fine also to experiment a bit with each before choosing.

<details>

<summary>I'll comment a little bit about Dialogue Manager and similar approaches</summary>

For an RPG I worked on, I similarly created a DSL for dialogues that designers could use. It was so similar to what's shown in the screenshot of Dialogue Manager that I had a moment of disbelief when I opened the repo, "wait, did I release this and forgot?" kind of moment.

(I found a very early iteration of the language design [story_examples.pdf](https://github.com/GDQuest/godot-open-rpg/files/12049931/story_examples.pdf))

But, it turns out, dialogue ties into everything in the game. As the designers worked, they asked for more and more functionality; being able to pop up an inventory; being able to give an item to the player; giving an item, but only if the player doesn't already have it; check if the player has enough encumbrance to accept the item, and if not drop it to the ground; start combat mode; start combat mode if some quest isn't done; trigger quests; trigger achievements...

As I added all those things, my DSL was starting to not be so small anymore, and becoming more and more generic. Taking a step back and squinting, I realized I had made Lisp all over once again (but with whitespace instead of parenthesis).

TL,DR: in the end, capturing this DSL in GDScript directly without the parsing and tokenizing step made things easier on everyone. Here's how it looks:

```gdscript
extends Dialog

var merchant := make_actor("merchant")
var wolf := make_actor("wolf")

func start_merchant() -> void:
	acting(merchant)
	say("hello, what do you want to buy?")
	say("I've got weares if you have the coin")
	if visits("start_wolf") > 1:
		say("I see you met the wolf, I'll make you a price")
		merchant.set_meta("special_price", true)
	say("do you want to talk, or buy?")
	choice("buy", "see_merchant_wares")
	choice("talk a bit", "merchant_talk")
```

From experiences, non-coders designers who used this had no trouble at all. They made less mistakes thanks to statistical analysis and overall worked just as productively.

The way it works is: declarative functions send out signals. UI captures those signals and displays the appropriate elements. Each visited function increments a counter, so things like `visits("start_wolf")` checks if the counter for the method "start_wolf" is higher than 0.

The benefits over a DSL:
- not having to maintain a parser/tokenizer/error messages/etc
- full access to everything: node tree, autoloads, UI, etc without any wiring
- you get in-editor autocomplete for game classes and objects
- can easily be extended to add new functionality, just subclass
- you get statistical errors in case of typos
- you can reach in and call any method of the dialogue if you need it for any reason (trigger a quest marker through alternate solutions; play a demo; ...)

I don't have access to that source anymore, but here's one old iteration packaged as a plugin (for Godot 3.1):

[godoversation.zip](https://github.com/GDQuest/godot-open-rpg/files/12049950/godoversation.zip)

As a side note, this also implements my architecture for quests, which was ~~stolen~~ inspired by [This talk by one of Inkle's co-founders](https://youtu.be/HZft_U4Fc-U?t=544). The way it works is as such:

Every time a node is visited, its counter is augmented by one. And that's it! That's the whole framework.

- Want to check if the player heard about the dragon? Check if `nodes_cache.heard_about_dragon > 0`.
- Want to know if the player harassed the merchant? Check if `nodes_cache.merchant_start_conversation > 50`.
- Want to track if such and such quest is on? `for quest in quests: if nodes_cache["quest_"+quest]: ...`
- Same for subquests

Each node also dispatches a signal when visited, so you can use that to pop toasters ("started quest so-and-so") or track achievements, etc.

Since a dialog is just a resource with methods, you can also create non-dialogue nodes to track events; eg, visit the node "wolf_killed" upon killing the wolf.

Finally my TL, DR about Dialogue Manager is: yea, why not, but I find this a little bit more cumbersome than just GDScript. Whereas Dialogic is very cumbersome, but is more balanced because it also offers an actual divergent way of treating dialogues.

Last note: If the GDScript API for dialogues works well, it's possible to imagine a simple regex-replace transpiler from something close to GDScript; eg:

`# start_merchant` → `func start_merchant() -> void:`
`@merchant: "hello, what do you wanna buy?"` → `  acting(merchant)\n  say("hello, what do you want to buy?")`

I'm not insisting on anything though, specially in the context of making a smallish demo, so if you feel comfortable with Dialogue Manager, by all means go for that.

</details>

food-please:
Thanks for putting the time into writing that, I really appreciate it. That's a goldmine of info. I enjoyed the example and the talk as well. 

Sorry that I've been so slow, school's much busier than I had expected. I am still working on it all, though!

ValorZard:
I’m excited to see this get merged in eventually. If you need any help with this, let me know!

food-please:
> I’m excited to see this get merged in eventually. If you need any help with this, let me know!

Thanks! I've just wrapped up an intensive two months' sprint, so I'm hoping to get back at this on Tuesday. I also am excited to see the final product!

mhuelreyes:
The Turn base Project is Replaced?

mhuelreyes:
> The Turn base Project is Replaced?

up

NathanLovato:
@mhuelreyes it's difficult to answer without context, but this is a Godot 4 remake of open RPG. Any other code or project you might be referring to is likely open-source thus will always be available. And this is a new open-source demo, for Godot 4.

mhuelreyes:
> @mhuelreyes it's difficult to answer without context, but this is a Godot 4 remake of open RPG. Any other code or project you might be referring to is likely open-source thus will always be available. And this is a new open-source demo, for Godot 4.

Im confused too. when I click the turn-base link in your YTvid its always redired it to this gitproject :(

mhuelreyes:
> @mhuelreyes videos can end up really outdated. You likely clicked a video that's several years old, and if so, the video is likely obsolete by now.

yhea,.. hopefully you will make an update godot 4 turn-base some day god bless thanks for your time :)

food-please:
Just a quick update because I'm having fun with working dialogue:

https://github.com/gdquest-demos/godot-open-rpg/assets/112480409/99d258d3-29d7-4a93-b9f7-614d3b1753f5

NathanLovato:
Thanks for sharing! Great progress there :)

food-please:
I've updated the OP with a revised feature list and continue making very slow (but sure) progress. Hopefully it's not too ambitious.

A quick question on design philosophy: I've got a series of objects (for example, battle results, interactions, various shops, etc.) that want to modify the player's inventory. Or perhaps it's a set of objects (for example, the combat state or an interaction) that want to change the music. In these cases there is a MusicPlayer or PlayerInventory somewhere far off in the scene tree. How would you propose going about this?

I could use an "inventory/music signal bus", except then ANYTHING could change the music/inventory. Which could be interesting or messy. I could also look at injecting dependencies but then you naturally have more tangled code. Should I be thinking of another approach?

Xananax:
All methods are good, as long as they are clear and fulfill the contract! Go with whatever you feel more comfortable with.

I'll still have a few thoughts below:

I personally do not have issues with "everything having access to data". My experience with that is that private/public and other methods to limit access are a hindrance. 

<details>

<summary>(open for more personal thoughts here)</summary>

> I like pseudo-private variables with a starting underscore (like in Godot), which tell you "you _probably_ shouldn't be using this, but hey, maybe you need it".
> 
> I understand that in very large organizations with many developers, one may feel the need to force people to not access some piece of data, but to my eyes, needing this at all is endemic and a sign you lack both documentation and code review (it also slows development tremendously while a developer waits a week for a higher-up to "unlock" the data they need through a PR).
> 
> I am also not against having globals, when they need to be globals. If everything needs access to a data blob (e.g, the inventory) and that blob is indeed always there, then I have no philosophical qualms with that. 
> 
> I am however not in favor of putting layers of indirection between dependencies and "pretending" the inventory isn't global, when it is indeed accessed by many objects that do expect it to exist and to be accessible. That is essentially a global, but obscured and made less clear only to make it look like it is not one.
> 
> For example, having a signal bus that talks to a single inventory is potentially just a global with extra steps.
> 
> Good reasons for not making something global:
> 
> - It is not guaranteed to be present (then making it a global would be making a false promise)
> - There are more than one of it (then it would be easy to mistake it with another)
> - Only few items have access to it (then making it global could be implying many things have access to it)
> - Logistics (for example, we want to be able to run a scene with a specific inventory for testing)
> 
> TL,DR: I try to only think of systems in terms of pragmatic use, with two main operational questions: What does this architecture "say" about itself? and: Is my current tooling requiring me to use such and such architecture for technical reasons?
> 
> I reject all "best practices" and "rule of the thumb" other than textbook cases. They apply in a lot of cases, but there are even more cases where they don't; in conjunction with other parameters (other systems with different requirements), they end up never applying at all.

------

</details>

In your case, would simply have globals and let systems talk to them. 

<details>

<summary>For everything that does not require access to the tree and no signals: I would use a class with static variables and functions</summary>

> In Godot 4.2, classes can be static and have variables, so true static classes are possible.
>
> 
> ```gdscript
> class_name GameData
> 
> static var PROFILE_NAME = "default"
> static var player_name := ""
> static var health := 10
> 
> static func save() -> void:
>     var file := FileAccess.open(get_profile_path(), FileAccess.WRITE)
>     file.store_var({
>       player_name = player_name
>       health = health
>     })
> 
> static func load() -> void:
>     if FileAccess.file_exists(get_profile_path()):
>         var file = FileAccess.open(get_profile_path(), FileAccess.READ)
>         data = file.get_var()
>         player_name = data.player_name
>         health = data.health
> 
> static func change_profile(profile_name: String) -> void:
> 	if PROFILE_NAME == profile_name:
> 		return
> 	PROFILE_NAME = profile_name:
> 	load()
> 
> static func get_profile_path() -> String:
>     return "user://profile-%s.data"%[PROFILE_NAME]
> ```
> 
> **caveat**: no signals can be dispatched from static classes
> 

------
</details>

<details>

<summary>For everything that requires access to the tree (such as music player): I would use an Autoload</summary>

> If I have multiple autoloads that depend on each other, I would nest them inside a parent Autoload (it's difficult to control the loading order of autoloads, but not that of an autoload's children).

</details>

<details>

<summary>For everything that requires serialization (such as inventory): I would use a resource</summary>

> I like to use resources instead of global classes whenever convenient, because they give me serialization for free, but also because I can explicitly require it in all items that require it
> 
> Like:
> 
> ```gdscript
> class_name GameData extends Resource
> 
> static var PROFILE_NAME = "default"
> 
> signal health_changed()
> signal died()
> 
> @export var player_name := "":
>     set(value):
>         player_name = value
>         save()
>         emit_changed()
> 
> 
> @export var health := 10:
>     set(value):
>         health = max(0, value)
>         save()
>         emit_changed()
>         health_changed.emit()
>         if health == 0: died.emit()
> 
> 
> ## Saves the current resource to disk
> func save() -> void:
>     ResourceSaver.save(self, ProfileData.get_profile_path())
> 
> 
> static func get_profile_path() -> String:
>     return "user://profile-%s.tres"%[PROFILE_NAME]
> 
> 
> ## Loads the resource. Godot caches calls, so this can be used every time needed.
> ## This function will create a new profile if one wasn't found
> static func restore() -> GameData:
>     if not FileAccess.file_exists(GameData.get_profile_path()):
>         var new_data := GameData.new()
>         new_data.player_name = "Player"
>         new_data.save()
>         return new_data
>     var data := ResourceLoader.load(GameData.get_profile_path())
>     return data
> ```
> 
> Then, in Player.gd, I can do:
> 
> ```gdscript
> class_name Player
> 
> var data := GameData.restore()
> 
> func take_damage() -> void:
>   data.health -= 1
> 
> func _ready() -> void:
>   data.died.connect(do_something_about_it)
> ```
> 
>  
> Or in UI:
> ```gdscript
> class_name HealthUI
> 
> func _ready() -> void:
>   GameData.restore().health_changed.connect(do_something_about_it)
> ```
> 

------

</details>

But, if you prefer not using globals, and passing dependencies explicitly, it doesn't need to be very tangled. You can export Node types, or Resources. For example:

```gdscript
@export inventory: Inventory
```
You can then either the player inventory to all items that need it.

If a node is created at runtime, then the factory would have an exported similar value, that it passes to the instances it creates.

The benefit of explicit declaration is no magic identifiers (though you do also get that with the Resource pattern described above), and easier to test (it's easy to pass a different instance, or to test an enemy on their own without having to also load the global).

food-please:
Thanks @Xananax for the response, as always. I am helped by both the philosophy and implementations.

cantdoright:
Hello, I would like to assist and help out however I can. 

Currently I have been looking over the latest commit on the repository and see a few things I can implement, but I also see in this pull request you have a few more things implemented that I can't access.

Would it be best to start on something tangential to your efforts or would you be open to collaborating on getting the event system done?

Xananax:
@cantdoright please note that there has been a long conversation to align vision and goals before getting to this PR, so it might not be so easy jumping in. Therefore, answering your question is difficult without more information. If you disclose details, it will help @food-please and me assess if your contribution would fit contextually.

I would favor you opening either:

1. An issue explaining your goals and intents
2. A PR based on this one (create a branch from this branch)

1 is better than 2 so you avoid doing any heavy work that wouldn't be pulled in.

Alternatively, if your changes are explained in a short text, feel free to comment here directly.

food-please:
> 1 is better than 2 so you avoid doing any heavy work that wouldn't be pulled in.

Just wanted to echo that this could be helpful. I'm hopefully on the tail end of this iteration of the milestone (which will, ideally, see aligned vision, etc.) but I'm waiting for a break from classes to wrap it up.

food-please:
I think that I've mentioned this before, but this update is a handful of re-usable scripts that are easily extended to create the different in-game events. I have a few things left to wrap up before submission, since I'll probably put off cleaning until later. I'm expecting to have a fake area transition and a puzzle, and that should be the demo.

My goal is to show off how easy it is to create puzzles/dialogue/etc., which I think was sorta buried in the previous demo.

secret_path

food-please:
Well, this is kind of exciting.

After ~6 months and dozens of hours of work I think this is ready for review. I have done a complete rewrite since [this point](https://github.com/gdquest-demos/godot-open-rpg/pull/212#issuecomment-1610113968), since I wasn't happy with the direction things were heading and the demo needed more context to make sense, I think.

I want to emphasize that this release serves to demo *Cutscenes*, and so the scenario would probably change to something a little more serious (and coherent?) for the final demo. I've tried to incorporate the feedback in the above conversations to produce code that is much more independent and maintainable and I'm pretty happy with how it's coming together.

***What is this release?***

Cutscenes!

- 3 main classes: Cutscenes (see OpeningCutscene at the bottom of the scene tree), Interactions (i.e. press space to chat/open chest/whatever), and Triggers (such as transitioning to a new game area).
    - Please note that I've approached these with the philosophy of Cutscenes plopping into the scene tree (and interacting with other nodes) in as "Godot-y" a manner as possible.
- 5 commonly used template cutscenes:
    - Conversations, implemented via [Dialogic 2](https://dialogic-docs.coppolaemilio.com/)
    - Simple pickups, which the player receives in their inventory when moving onto the same cell.
    - Treasure chests. They can be closed afterwards, which I'm pretty happy about.
    - Area transitions, which smoothly transport the character to a new game area. **Please note** that this currently happens within the same scene. I expect that we will want a "world" update of some sort that deals with loading/freeing areas and moving the player to matching "portals".
    - Lockable doors, which are a complex scene created from a variety of triggers and interactions.
- A variety of inherited scenes to show how specific game behaviour may be derived from Cutscene/Interaction/Trigger and implemented quickly. See, for example, the StrangeTree in the town or the WandPedestals in the house.

3 small areas to explore! Please see the note under "area transitions" above. This model will change in future releases.

A variety of small classes:

- A WIP inventory. The overarching model (and manner of accessing the inventory via code) will probably stay the same, even as the implementation changes. At the moment, it serves to demo Cutscenes and so I thought it made sense to include it as a "first pass".
- A simple music player
- Screen transitions (both simple fade and texture based)

There are a few minor tasks to be done before merging with main: I would like to change the app icon, mouse input may need to be adjusted slightly, and I thought I would add a few comments where there is confusion. I'll be able to better do these things after review, however.

I know that you're all crazy busy, but I hope that you enjoy playing through the demo and that there is not too much code to go over. I'm looking forward to the combat update, and I think this gives a really good launching point.

maps/house/wand_pedestal_interaction.gd
```diff
@@ -0,0 +1,57 @@
+@tool
+
+class_name WandPedestalInteraction extends InteractionTemplateConversation
+
+signal colour_changed()
+
+#TODO: will be replaced by proper inventory items (w/ sprites & icons) during the inventory update.
```
Xananax, line 7: I don't really have issue with keeping these as they are. Often items like these either:
1. Are a one-off in one dungeon
2. Or only work with a specific item or set of items

So there's no point having code lie and pretend it's generic. If it could be generic at no complexity cost, that'd be ok by me; 
Concretely, often you'll need specific item placement or even maybe different sprites depending on the item.
An enum is completely ok here; it filters the kind of item the pedestal can receive.

(it's also fine to make it generic -- just potentially not necessary)

maps/house/wand_pedestal_interaction.gd
```diff
@@ -0,0 +1,57 @@
...
+const ICONS: = {
+	"red": preload("res://assets/items/wand_red.atlastex"),
+	"blue": preload("res://assets/items/wand_blue.atlastex"),
+	"green": preload("res://assets/items/wand_green.atlastex")
+}
+
+@export var wand_placed_timeline: DialogicTimeline
+
+var colour: = ""
```
Xananax, line 16: I am not sure how Dialogic works and if it can deal with types that aren't strings, but if it could, is it possible to skip the item -> string -> item conversion and just pass the `ItemType` around?

If you intend to replace items by resources, then this is only as useful as the possibility of passing the actual item ref later (or a constructor for it).

maps/house/wand_pedestals.gd
```diff
@@ -0,0 +1,47 @@
...
+	var inventory: = Inventory.restore()
+	inventory.item_changed.connect(_on_inventory_item_changed.bind(inventory))
+	
+	_blue.colour_changed.connect(_on_wand_placed)
+	_green.colour_changed.connect(_on_wand_placed)
+	_red.colour_changed.connect(_on_wand_placed)
+
+
+func _on_wand_placed() -> void:
+	if _blue.colour == "blue" and _green.colour == "green" and _red.colour == "red":
```
Xananax, line 21: Instead of checking a string value, since the intent is only to check if the item is valid, maybe an `is_valid` or `is_activated` boolean works better and communicates intent better.

maps/house/wand_pedestals.gd
```diff
@@ -0,0 +1,47 @@
...
+		# to wait to update the pathfinder until then.
+		# The first frame waits until the "spikes" blocking area has been freed. The second frame
+		# waits until the physics engine has registered the change in collision objects.
+		_spikes.queue_free()
+		await get_tree().physics_frame
+		await get_tree().physics_frame
+		FieldEvents.terrain_changed.emit()
+
+
+func _on_inventory_item_changed(item_type: Inventory.ItemTypes, inventory: Inventory) -> void:
```
Xananax, line 38: Why is this not in each pedestal's logic? It's fine to have it here, but at the moment, logic is split in two places; I would recommend either controlling the inventory and dialogic from each pedestal (inventory checks, removals, additions; dialogic removals and additions); or controlling both from the pedestal controller (in which case the pedestals act as areas and contain no logic other than presentational).

maps/town/strange_tree_interaction.gd
```diff
@@ -0,0 +1,33 @@
...
+	super._ready()
+	
+	# We want to open the secret path once the player has performed a specific action.
+	# The easiest method is to allow Dialogic to determine when this should happen. This could be
+	# done via the EmitSignal event, the Call event, or, as we've opted to do here, an in-dialogue
+	# signal.
+	# Note that the signal argument must match exactly for the lambda to play.
+	Dialogic.text_signal.connect(
+		func(argument: String):
+			if argument == "clear_secret_path":
```
Xananax, line 18: I'm not managing to get dialogic to run, so I'm not sure what is best to do here; but it seems to me that this is a double opportunity for typos/mistakes that will silently fail.
You could mistype the signal on the side of Dialogic; you could mistype the string array here.

I would recommend one of the following:

1. Runtime check: Force the dialog to always send a signal. Then you can do a match against the argument, and if a signal is not recognized, pop an error (``null` could  be a valid argument, so that'd be a one line change; `assert(argument == null, "noooo")` at the bottom). I don't know if Dialogic always sends an event, if not, this isn't really a guard.
2. Other runtime check: read the global `StrangeTreeExamined` var off Dialogic
3. Static check: have an internal counter increment every time the dialog is triggered (that implies switching the Dialogic dialog resource from code instead of having Dialogic track the number)

maps/town/fan_interaction.gd
```diff
@@ -0,0 +1,48 @@
...
+
+func _ready() -> void:
+	super._ready()
+	assert(_adoring_fan, "Gamepiece was not found, check the node path!")
+	assert(controller, "Controller was not found, check the node path!")
+
+
+func _execute() -> void:
+	match _quest_state:
+		1: timeline = TOKEN_CHECK_TIMELINE
```
Xananax, line 23: This is inconsistent with the behavior in the tree interaction. I think either or should change to follow the same logic (unless there's a reason for doing it differently here that is escaping me)

Xananax:
Very nice! I played the demo, the dialogues are fun. It could stay like this, with the 2nd degree and programmer in-jokes, I think it works perfectly well for a demo.

From a quick overview it seems like I wouldn't have much to change, I'm mostly asking questions as you'll see.

There are areas that I might ask you to change structurally, mostly for the sake of simplification, but I am not sure the simplification is truly simpler without having a more complete picture, so doing it now may hinder progress or force roundabout solutions. I would rather do a simplification pass at the end.

maps/house/wand_pedestal_interaction.gd
food-please: I'm happy with less code. An alternative would be some sort of a template that could check to see if a given item is on top of it, but that seems pretty niche.

maps/house/wand_pedestal_interaction.gd
food-please: I've pulled out the match statements and added a handful of export variables that should, hopefully, clarify what's going on in both the code and the scene tree. You can specify which colour is expected and these values are easily converted to their corresponding item id. Also, everything occurs within a single GDScript file and a handful of nodes in the scene tree. Not sure what you'll think of current architecture, so I'll leave it unresolved until you've gone over all these changes.

So, Dialogic 2 claims that it can access autoloads/globals/etc. within events, but I wasn't able to get it to work. No matter what I did, I would get a string object back from the signal. I'm not sure if I'm just not using it correctly or what the deal is. The happy medium at this point is that I've changed Dialogic to use the Inventory.ItemTypes enum keys, so everything is measured by that single enum (which acts as a sort of inventory database at this point).

I think this is basically resolved at this point. I hadn't been planning on keeping this puzzle once we add a "world" to move around in, but I suppose it might still be useful in some form.

maps/house/wand_pedestals.gd
food-please: This script is gone, and everything is tracked within wand_pedestal_interaction.gd, where this is all tracked by a dictionary of static booleans (one for each pedestal, dynamically created and evaluated. Theoretically you could have a hundred pedestals).

maps/house/wand_pedestals.gd
food-please: This all happens in a single file now (please see the conversations above), so this particular issue is resolved. 

However, I'm **not** happy with the current implementation that has Dialogic-specific variables that must be updated to match the player's inventory. Dialogic should, somehow, be able to check in-game variables from timelines. I just haven't figured it out yet, I'll see what I can find out on their Discord.

maps/town/strange_tree_interaction.gd
food-please: Agreed with analysis of the issue. It was a clunky solution anyways. The only reason I had done this was to show off in-text Dialogic signals (rather than the events), but the trade-off wasn't worth it here.

It's also apparent to me that Dialogic signal events are pretty common in conversations, so I've added them to the default conversation template (which means a derived class now only needs to implement the _on_dialogic_signal_event callback. It's now already hooked into the interaction when deriving from the template). This has simplified the strange tree interaction and it now follows the same pattern as other interactions.

maps/town/fan_interaction.gd
food-please: I've tried to simplify this interaction. I think I was tired when I wrote the first draft.

There's now a single timeline (so, no switching timelines on the fly) where most of the logic happens via an integer-quest-status-variable. The only unusual thing that happens here is that the script reads the Dialogic variable to identify the first encounter with the NPC, since that's the one where we want him to run into a corner and wait for the fetch quest to finish.

But it has been reduced to a single interaction and a single timeline, so hopefully it's much simpler to follow. I'll resolve this for now and rework it if you think that it's still too obtuse.

food-please:
> Very nice! I played the demo, the dialogues are fun. It could stay like this, with the 2nd degree and programmer in-jokes, I think it works perfectly well for a demo.

Thanks! I had fun filling out the little town. I think I would like to make it more JRPG-like for a future demo, but that's a bridge to cross when we get there (hopefully before Godot 5).

> From a quick overview it seems like I wouldn't have much to change...
> There are areas that I might ask you to change structurally, mostly for the sake of simplification... I would rather do a simplification pass at the end.

Feel free to let me know whatever you think could be simplified. I find that, more often than not, when I explain myself I find out that my reasoning wasn't very good. I've really enjoyed some of the more simple paradigms that have come out of this project (I like the way the inventory works! Looking forward to a save/load update).

> Note: I couldn't make Dialogic work in the editor (it ran in the game without problems). I'm not sure why. It isn't very important for now, but there may be subtleties I missed because of that.

That's a deal breaker. I check it before writing you here and it works on my end. I clone the project and open the project file once. It throws a pile of errors. Once it's fully loaded I close it, wait a sec, and reopen the project. This time there should be no errors (though there is this pesky rid bug that I haven't been able to get rid of quite yet), and if you look in the project manager the Dialogic plugin should be enabled. At that point, I can see the "Dialogic" tab in the tab group at the top (i.e. "2d" "3d" "Script"... "Dialogic"), and clicking that opens up the Dialogic manager. I haven't fully figured out all of Dialogic's features yet. Hopefully this is a good start.

I think that I've resolved all the points above, so I'll wait to hear back from you on how to go forward.

Thanks, appreciate all the feedback.

One question: I'm not sure that mouse/touch input works well for this project. I find navigating the field to be a little clunky. This kind of movement seems, to me, to best suit controllers/keyboard. Do you think we should try to maintain mouse support?

NathanLovato:
> One question: I'm not sure that mouse/touch input works well for this project. I find navigating the field to be a little clunky. This kind of movement seems, to me, to best suit controllers/keyboard. Do you think we should try to maintain mouse support?

Mouse and touch can work very well for this kind of game, it's just a matter of adding a couple little things. You can leave it as-is for now and we can take care of adding the missing bits later if you'd like.

Regarding Dialogic, same issue as always with many Godot plugins: on first load of a project they fail to run as their code runs before project resources have been imported. It's okay, it's a Godot limitation, and it's going to be an issue with any dialogue editor you would use. It can be worked around, but it's generally not worth it.

I tested and everything's working really well, great job! As Jad took care of the last code review, I'll let him check your changes and merge, unless either of you would like me to dive into the code.

food-please:
> Mouse and touch can work very well for this kind of game, it's just a matter of adding a couple little things. You can leave it as-is for now and we can take care of adding the missing bits later if you'd like.

Sounds good, I've got it pretty much working for the moment. I think that I would prefer to iron out the mouse related bugs (there are a couple, one of which stems from Dialogic's input handling) in a later pull request, since many of these are related to the current demo. I expect the scenario to change between now and the "final" demo, making some of these disappear.

> I tested and everything's working really well, great job! As Jad took care of the last code review, I'll let him check your changes and merge, unless either of you would like me to dive into the code.

Thanks. I'm in no particular rush, so please don't be inconvenienced on my account. I'm more than happy to wait.

NathanLovato:
Our launch has been hectic, so as much as I wanted to review this PR, I haven't had the time. I'm sorry for the delay. I'll try to get to it by the end of the weekend.

NathanLovato:
I could finally get to it! Great job overall, everything is clear and seems pretty composable and flexible already, perfectly adapted to this kind of game demo.

This is good to merge for me. Like Jad said, it would make more sense to do a review and refactoring pass at the end, when all systems are in place and all requirements and interactions between every system are settled (field code may need some additions and tweaks when adding character stats and abilities, for example). Otherwise, we might end up doing some unnecessary work. And I find the code pretty easy to change and refactor, so I'm not worried about that.

Thank you so much for your contributions!

PR #221, opened 2025-02-14
Player Battlers & Actions

food-please:
Another PR looking to address the [combat issue](https://github.com/gdquest-demos/godot-open-rpg/issues/215#issue-2127576871). Almost there!

The following PR adds player agency to combat. Please note that I have incorporated some elements from the 2D Secrets course while deviating significantly from it in other ways. The main change is that I envision players queuing actions for their Battlers, rather than only selecting an action once their Battler has fully charged. This is mainly to give the player something to do in the downtime which, in my opinion, was "wasted space".

This leads to the following combat loop:

- The player selects a battler from their list of available battlers. If a player battler is given an 'AI' controller, they cannot issue orders to that Battler, though they will be teammates and live/die together.
- On selecting a battler, the player is then given a list of actions from which to choose one. If an action has no valid targets (e.g. raise dead, when there are no friendly dead battlers), it is still displayed but is unselectable.
- On selecting an action with some valid targets, the player may then choose a valid target from the list provided by the action definition. Actions may specify whether or not they target friendlies, enemies, or both and may choose from targeting self, targeting a single Battler, or targeting all battlers of the chosen team(s).
- One selecting a valid target, the action is cached and will be played out once the Battler's ready_to_act signal is emitted. Note that only one action will be played at a given moment.
- The player should be able to go forward or backwards through the menus easily in order to select the optimal action/target combination.

Please try out the combat examples (largely unchanged from the previous demo. Visit the two ghosts.) and let me know if you feel that the workflow is (un)intuitive and easy to navigate. I'm pretty happy with how it's turned out.

The other major deviation I've taken from the 2D secrets course is that I wanted to remove any yield/await calls on user input. That is, it would await for the player to select an action, and you really couldn't do anything else at that time. Breaking out of the loop (e.g. ending combat) would end up with strange yield-based errors. The alternative, of course, is signals, so hopefully this is a little simpler to follow.

Sorry that this took so long. There were many edge cases to work through; the whole real-time combat with discrete targets was much more complicated than I expected. There may be some architecture/style issues as always.


food-please:
ui_progress

food-please:
The bear looks darned silly, but that's the way it is until tackling gfx.

Also, I should note that Battlers are not currently target-able via mouse. That's a fix for another time, I think.

NathanLovato:
I just took some time to test and read the code. It works very reliably, it is clear and readable as usual.

UX-wise, to me, the idea of letting the player choose character commands anytime is not intuitive right now. I don't see it working too well given that the wait between actions in this kind of game is not very long.

Actually, for future users of the repository, I'd steer the project towards more classical visuals and either:

- A classical ATB model (with gauges per character, based on older final fantasy games)
- A fixed turn-based model, perhaps with the turn order displayed

Just so that it offers a more straightforward reference to users, as ultimately this is a demo for learning purposes.

For now though, no need to worry about that. It's more than good enough as it is, and code-wise either of the variations above is actually really close.

You did a great job once again, and I want to thank you deeply for your work! Let's get this merged.

PR #225, opened 2025-06-24
chore: update readme godot version

imLymei:


food-please:
Good catch, thanks!

food-please:
Just a follow up thought, do we even need that line in the readme? The expected Godot version shows up in the project manager, and we'll be changing it whenever a new stable version comes out (though 4.5 looks enticing, maybe a solid RC?).

PR #229, opened 2025-09-16
Upgrade to Godot 4.5 & Dialogic 2 Alpha 16

food-please:
A quick maintenance update, primarily addressing [this issue](https://github.com/gdquest-demos/godot-open-rpg/issues/227). This PR does the following:
- Upgrades Dialogic 2 to Alpha 16, which introduced a number of improvements (and changed portraits slightly)
- Upgrades Godot to the current release, version 4.5.
- Ensures that Dialogic works of key/button releases, rather than presses. Otherwise, the player would not be able to escape a dialog - it would begin as soon as it ended.

I could not reproduce the issue (see the link above), but hopefully this heads off any other issues that developers might encounter.


NathanLovato:
Thank you very much, as always!

PR #230, opened 2025-09-19
Draft turn-based combat core

food-please:
I'm trying to keep to short PRs, and this is my first attempt. Please note that this introduces code that does not yet replace the old logic, so the PR may be too short for your liking.

We've discussed re-writing the combat to align with traditional turn based titles (e.g. Dragon Quest, Final Fantasy I, etc.). With that in mind, I've proposed the following structure:
- actor.gd : An actor can be anything (usually a Battler, but could also be an angry volcano that spews lava every 3 combat rounds, for example) that gets to take 1 turn per combat round. An actor replaces the concept of CombatAI and PlayerControllers. It decides which actions the Battle will perform based on computer or player logic.
- turn_queue.gd : Handles who is the current actor and, after all actors have taken a turn, moves everything along to the next combat round.
- battler_roster.gd: Will replace BattlerList. It is more focused, not dealing with the logic of who's turn it is (since that's the TurnQueue's domain).

So, this should replace the active turn queue system. Next steps are as follows:
- integrate the new classes into the existing framework. That means pulling out CombatAI & ActiveTurnQueue, then hooking up the player action selection to the new TurnQueue signals.
- remove the old code.

If this short PR format works, I'll implement the above in two shorter PRs. Otherwise, I'll do the whole thing in one fell swoop. Just to reitereate, the demo does not currently make use of the turn-based framework, it is only included for now under src/combat/actors.

Note: the [second commit](019b51c20714d0e00a3fa6b77f35dea9bb9e7e2a) includes a pile of .import files that had odd little changes. **_However, the commit also includes changes to combat.gd and battler.gd which should be included._** I'm not sure if you want to bring all of those changes over into main, so I thought that I would check first.


food-please:
Thanks @NathanLovato, this is great. Sorry that I didn't drop the PR size earlier, probably would have saved a pile of revision time.

Regarding complexity & design, I really appreciate when you identify paradigms (in this case, pulling from the CombatActor group) that make it harder to understand the code. I think the core goal is simplicity, and I get sidetracked so quickly. In this particular case, I was imagining that polling the group was an easy way to address CombatActors being added/subtracted from the battle (i.e. a sort of "call reinforcements"). Of course, on reflection, that is not a necessary feature for the demo.

So, please feel free to let me know when you have any thoughts on what could be simplified!

NathanLovato:
I will! I just can't say much until more pieces fall into place and we can see or know why every piece is there (in the context of the playable system).

PR #235, opened 2026-01-13
Change combat system from ATB to turn-based

### LINKED ISSUE #215: Combat Milestone
## Feature implementation: v 0.3 - JRPG Combat ## 

I'm hoping that we can create a **simple** (J)RPG combat system, similar to what is outlined in the [2D Secret's course](https://school.gdquest.com/courses/godot_2d_secrets_godot_3/jrpg_combat_core/introduction).

I think the goals should be as follows:
- Transition smoothly to and from the combat state, keeping player/npc/etc. position on the overworld in place.
    - Allow combat in the middle of a dialogue. A prototypical example would be a boss battle that starts with unnecessary villanous pontification and ends with the villain running away, tail between legs.
    - Transition to combat typically from contact with overworld monster NPCs. Clean up the NPCs on player victory.
- Keep a persistent player party between combats. Allow healing on the overworld to demonstrate persistence.
    - This includes battler stats, as mentioned in this [overlapping issue](https://github.com/gdquest-demos/godot-open-rpg/issues/213).
- Allow for a variety of enemy formations, including changes to music, background, enemy types, etc.
- Include combat results/game over screens as necessary.

**Implementation and Scope**
TBD. Will edit once I know others' vision for project. Basically, this will describe the model we want (what stats to include, what visuals are needed, real-time/turn-based, etc.).

**Discussion and Feedback**
So, we've mentioned this tangentially in other issues/PRs, but I would like to nail down the combat system. In my mind, I think it's a good idea to implement the prototypical combat model, perhaps with a simple resource system (Action pointss, as in the 2D Secrets course?).

Do you have any particular thoughts on combat design? I would find it helpful, I think, to know what the GDQuest folks would like the final product to look like. We can re-implement the 2D Secrets course, or we can do something entirely so that folks visiting the repository can look at multiple combat implementations.

Speaking of appearances, I'm running out of free art and will probably need to start drawing. It would also be good to hear of your thoughts for project visuals. I've included a few mockups below, not sure what you think of any of them. I remember reading somewhere that you're preferring vector over pixel art, which I'm able to work with. I know that you also were using Krita-painted assets, which I could revisit as well.

tumblr_plh1ldRqIn1svbgupo1_1280
Pixel-boy at https://2pblog.tumblr.com/image/182081376051. Retrieved 2023-09-23.

viktor-stambert-game-mockup-rpg-frontview-done-fixed
Viktor Stambert at https://veo_the_artist.artstation.com/projects/v28BEx. Retrieved 2023-08-02.

rpgremake2
junkboy at https://pixeljoint.com/pixelart/52968.htm. Retrieved 2022.

food-please:
This should close [the big combat issue](https://github.com/gdquest-demos/godot-open-rpg/issues/215). I've added some notes here that hopefully explain the thinking behind some of the design decisions. As always, I love to hear about/see alternative ways to implement things.

The goal of this PR is twofold:
- Transition the combat system to be a sequential, turn-based system such as Final Fantasy 1, Dragon Quest 1, etc. The player issues orders to the characters, who then trade blows with the enemy battlers.
- Simplify the code in a number of ways:
   - Architecture
       - **Simplify architecture**, since I was tired of tracing combat logic across 4 or 5 scripts. Now, pretty much everything combat-logic-related happens in the Combat script, which sets up and shuts down the scene as well as walking through Battler turn-taking. BattlerList/BattlerRoster is now just the ancestor of all Battlers and looks them up by type (so, no NodeGroups, references to other (grand)parent nodes, etc.). Pretty much everything UI related should be found under CombatUI. Hopefully it's much more straightforward.
        - **Removed the turn bar**, since we've moved away from the ATB model. While it's relevant in turn-dependant games (e.g. Octopath Traveller), it's more complex than what is needed for simpler RPGs.
        - **Removed CombatActor** components. Their functionality is now simplified and included in Battler (which had a number of redundant methods with CombatActor anyways). Consequently, all active Battlers will get a turn in combat.
    - UI
        - The **combat UIs are no longer derived from a template base menu** class. Rather, they are their own thing, with simplified functionality (e.g. the player party display simply displays player stats, the action menu doesn't share functionality with other menus, etc.). This means that the overall amount of code is smaller and devs don't need to open several files to see what's going on.
        - The **combat UI is no longer embedded in the CombatArena** template scene, making arena design (should be mostly VFX, SFX, and the enemy party anyways) simpler and making the combat UI easier to find in the overall tree.
        - The **combat UI is now animated by a single animation player**, rather than each individual element responding to a myriad of signals. Additionally, the fade in and out and combat beginning and end are now handled in the same place as the rest of the combat start/end logic (the combat node).
        - **Combat menus no longer fade in and out**. Response times should be snappier and there's no need to inflate and convolute the code by await-ing for menu animations.
        - I **removed Battler and BattlerRoster references from the UI**. The CombatUI seems unnecessarily complicated, so I tried to reduce dependencies. UIActionMenu no longer stores references to anything other than a list of actions. The UITargetCursor no longer stores references to anything other than a list of targets.
    - A note on Battlers & BattlerActions
        - Related to the above, actions need to know about who is the source of the action and who can be targeted by the action. Previously, the source Battler and BattlerRoster were injected into the action as method arguments. Instead, now we **cache references to the source Battler and the BattlerRoster within BattlerActions**, which means that the references don't need to be tracked by externals systems (the UI, for example). A little more complexity in the action (it requires a few parameters to be setup and will fail if the references are not set when its methods are called), but I think it results in much simpler logic where each UI system can focus on exactly what it is supposed to do.
        - Probably warrants further discussion, but I've been trying to figure out what exactly BattlerActions should be. They're a resource, currently, so that names, icons, descriptions, etc. can be set via the editor. However, behaviour is implemented by logic, so they must be necessarily implemented (**I made BattlerActions abstract**, since the default interface does... nothing). This means that there's some redundancy, since the resource must be saved as a .tres file to modify the exports. An additional benefit of resources is the ability to integrate a [plugin that allows modifying resources in a table format](https://godotengine.org/asset-library/asset/1479). However, the script is also necessary (is there a 1 to 1 correspondence between actions and their saved resource? Are BattlerActions ever shared between Battlers? Probably with player Battlers.) Anyways, I'm not sure if there's a more elegant/simple pattern to be used here.

Please note that there are many commits as well as notes above. Hopefully you'll find that this PR is fairly easy to work through, since I've tried to mostly remove and re-organize into something more cohesive.


NathanLovato:
Thank you for the kind words. I'm just wrapping up the last day of work after launching the 4.6 update of the courses.

I tested this first and foremost and it's working well. Thank you again for the work.

Then in terms of the structure, like before, I'd need to really take the time to work with the code to be sure. It looks fine, it's still the same kind of approach as before, with a nice simplification. I'm always wondering about, for example, having an abstract resource base class to represent actions VS having instead a larger script with a simple kind of composition where you would have a library of functions and actions would just take a pointer to a function to represent an action.

But because Godot is pretty editor-centric, generally resources make sense as they are editable in the inspector out of the box.

The main thing I'm thinking now is that probably designers and developers would want every skill in an RPG to have possibly unique logic (play a particle effect, spawn an animation sequence, attack 3 times in a row, dealing damage each time, heal allies but not self, etc.). And the simplest way to do that is to generally code every skill individually, in a function.

Using resources and inheritance probably makes it harder in practice for a solo developer to have that flexibility - you need to make new kinds of resources often. And once you've used resources and inheritance everywhere, it's a bit tedious/time-consuming to replace. This is among the sorts of things I'm thinking about, though I'm not set on any of that.

NathanLovato:
Thank you very much for all the work! It's a good simplification already.

There is more that I would like to do but going further necessarily involves working with the project from a design perspective for some time.

There are some code details like for example, the stats, we refer to them by strings. They could use an enum instead. The combat AI tries to select an action in a loop. The algorithm should probably not involve a loop that can fail to decide something. There are some implementation details like these that could be cleaned up.

But for more important questions that like actions and whatnot, to answer that you're better off actually building a little complete game demo first, nailing the mechanics and content, and then reworking the code based on what those needs. With the actual final graphics and all. This would probably be what this project needs next.

Thanks again for all the time and energy you put into this! The community will love it.

PR #237, opened 2026-04-28
Update Godot to 4.6.2 and Dialogic to snapshot from April 27, 2026

food-please:
A quick maintenance update, addressing the issue that Dialogic was out of date and throwing errors. Additionally, Dialogic has added an in-engine updater that should make it easier for users to update the plugin on their own. 

This PR also modifies Dialogic to work off key-releases, and mitigates a bizarre timeline_end feature that produces an empty dialogue panel if left alone.

This should resolve previous lingering warnings that seemed to be produced from using the older engine and plugin versions. Conversely, Dialogic is currently leaking memory and the editor produces a warning when closed. This seems to be related to a "polyphonic" audio stream player that isn't being freed correctly.


NathanLovato:
Thank you very much!

